Issue #56, opened 2026-07-11
0.11.1 Core dump on wayaland


alfixss:
OS + version: Arch Linux
Dusk build: 0.11.1
Audio backend: PipeWire
What you did: try to launch the downloaded binary (without installation)
Expected vs. what actually happened: results in "core dump"

```
dusk-studio-0.11.1-Linux-x86_64: ./DuskStudio/DuskStudio 
Aborted                    (core dumped) DuskStudio/DuskStudio
```

Notes: 
- tested with different compositors (sway, niri, mango, labwd), same behavior.
- on the same system, but on a x11 session (openbox) the app launches OK.

alfixss:
Hello Mark, I've checked the 12.1 build and the app launches perfectly on wayland. Thanks!

marc-korte:
@alfixss Awesome! 

The issue was totally on my end. The wayland-JUCE fork I am using for local builds was not the one the new versions were using for the CI builds, but the 0.9.0 version used it. I have that fixed going forward.

There is a mixdown bug on 0.12.1, I caught it last night and (hopefully) have it fixed. Try 0.12.2 when you get a chance and let me know if you run into any issues. I plan to post an Patreon update today, I've been slacking lol.

Marc

PR #60, opened 2026-07-11
0.12.0 release prep: JUCE pin standardization, Wayland launch fix, workflow hardening

marc-korte:
Prepares the 0.12.0 re-tag and fixes the #56 root cause.

The crash (#56): releases since 0.11.0 ship the hybrid-Wayland JUCE fork, whose startup handshake aborts on wlroots/smithay-family compositors before any window exists (registry binds at compositor-advertised versions, libdecor init). 0.9.x shipped stock upstream JUCE and worked everywhere as a plain XWayland client. Since Dusk Studio forces every window to an X11 peer, the Wayland connection was never used for windowing. Main.cpp now hand-rolls the JUCE entry point and sets JUCE_XWAYLAND by default on Linux — the fork's own skip path — restoring 0.9.x behavior on every compositor. DUSKSTUDIO_NATIVE_WAYLAND=1 re-enables the hybrid backend for development. (The earlier preflight from #59 still covers genuinely-missing X displays.)

JUCE standardization: dev, CI and releases now build one JUCE. All Linux CI jobs (build, tests, TSan, Raspberry Pi) clone the pinned dusk-audio/JUCE-wayland mirror snapshot instead of upstream 8.0.4 plus an inline SIMD patch; the release pin bumps to dusk-wayland-v2, content-identical to the dev fork HEAD including the previously-uncommitted ALSA endpoint seeding and XEmbed defer-mapping fixes; bump-version.sh gains an anti-drift gate (dirty fork fails, drifted tree warns that a new mirror tag is due).

Release plumbing: donor plugins clone pinned to a SHA like the JUCE pin; a fail-fast token preflight on tag runs (the v0.12.0 attempt burned a 17-minute green build on an expired RELEASES_REPO_TOKEN); the publish step distinguishes the benign sibling-arch race from real failures instead of masking both. appdata + CHANGELOG entries refreshed for the re-tag covering the July production wave and the newer work.

Verified: 364/364 tests, Xvfb selftest 26 PASS / 0 FAIL, dead-display launch prints guidance and exits 1, all four Linux workflows converted and YAML-validated, fork HEAD tree verified identical to the v2 snapshot.

## Summary by CodeRabbit

* **New Features**
  * Linux now defaults to XWayland for improved compatibility, with an option to enable native Wayland.
  * Added clearer guidance when launching without XWayland.
  * Updated release information with console EQ improvements, MIDI modernization, workflow and session enhancements, and other 0.12.0 changes.

* **Documentation**
  * Updated third-party licensing references and release metadata.
  * Corrected the 0.12.0 release date to July 11, 2026.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -47,61 +47,34 @@ jobs:
             libwebkit2gtk-4.0-dev libglu1-mesa-dev xvfb \
             libwayland-dev libxkbcommon-dev libdecor-0-dev
 
-      - name: Clone JUCE 8.0.4 (upstream)
-        # CI uses upstream JUCE. The plugdata-team wayland-juce8 fork
-        # carries 5 local commits (XEmbed mapping, SIMD<int64> alias,
-        # Wayland peer-creation latch, etc.) that exist only on the dev
-        # box — not pushable. The runtime-only fixes (XEmbed, Wayland
-        # peer) don't matter for CI; the SIMD<long long> alias does
-        # because GCC on x86_64 instantiates SIMDRegister<long long>
-        # for jmax<juce::int64>(). Patch it in below.
+      - name: Clone JUCE (Dusk-owned Wayland mirror, pinned)
+        # CI builds against the SAME Dusk-owned mirror snapshot the release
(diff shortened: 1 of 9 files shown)


marc-korte:
@coderabbitai review

PR #85, opened 2026-07-16
De-JUCE: OOP plugin hosting MIDI path onto dusk::MidiBuffer

marc-korte:
Completes the hosting-layer MIDI migration: no `juce::MidiBuffer` remains in either the in-process (#84) or out-of-process plugin path.

`RemotePluginConnection::processBlockSync` took a `juce::MidiBuffer` for the out-of-process round-trip. It already serialises MIDI to raw bytes for the SHM transport and rebuilds the reply the same way, so `dusk::MidiBuffer` is a drop-in. The input loop reads `meta.numBytes` / `meta.data` directly; deserialising the child's echoed MIDI `addEvent`s raw bytes instead of constructing a `juce::MidiMessage`.

`PluginSlot` keeps `juce::MidiBuffer` (the in-process JUCE instance path needs it) and bridges the block into a pre-sized `dusk::MidiBuffer` scratch before each OOP call; the child's MIDI output is discarded by the caller as before (PluginSlot reads only `readOutChannel` after the call).

`RemotePluginConnection.h` goes fully JUCE-free (de-JUCE gate 199 -> 198); the `.cpp` stays coupled via `juce_events` (the `ParamChangedFromChild` async dispatch, unrelated to MIDI).

The ipc-stub round-trip test now feeds a note-on and asserts it echoes back byte-exact through the real child process (`--ipc-stub` echoes midiIn -> midiOut), covering the serialise/deserialise flip end-to-end.

Verified: app builds with zero new warnings, ctest 383/383, gate 198, Xvfb selftest 15/15.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MIDI handling during remote plugin processing for more reliable event transfer.
  * Reduced unnecessary per-block memory growth during remote plugin operation.

* **Tests**
  * Expanded IPC round-trip validation to verify MIDI event data and timing positions.
  * Updated MIDI-related self-tests for improved transport coverage.

### DIFF
--- a/src/engine/PluginSlot.cpp
+++ b/src/engine/PluginSlot.cpp
@@ -342,6 +342,10 @@ void PluginSlot::prepareToPlay (double sampleRate, int blockSize)
     // a stereo-only plugin is in the slot.
     stereoScratch.setSize (2, preparedBlockSize, false, false, true);
 
+   #if DUSKSTUDIO_HAS_OOP_PLUGINS
+    oopMidiScratch.reserveBytes (duskstudio::ipc::kMidiBytes);
+   #endif
+
     blocksSinceLoad     = 0;
     consecutiveOverruns = 0;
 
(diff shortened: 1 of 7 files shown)


marc-korte:
@coderabbitai review

PR #89, opened 2026-07-16
De-JUCE device layer Phases 1+2a: dusk::audio device interfaces + DeviceManager seam

marc-korte:
First two phases of replacing `juce::AudioDeviceManager` with a bespoke device layer. **Additive** — nothing wires to it yet, so no behavior change. Phase 2b flips AudioEngine onto the seam; Phase 3 swaps the backing to native and unlinks `juce_audio_devices` on Linux.

## Phase 1 — JUCE-free interfaces (`src/engine/device/`)
- `ChannelSet` — 64-bit channel activation mask replacing `juce::BigInteger`; `setRange`/`count`/`highestSetBit` mirror the ops the device code + selector use.
- `IODevice` / `IODeviceType` / `IODeviceCallback` (+ `CallbackContext`) — the device, factory, and RT-callback interfaces, all `std::string`/`std::vector`/`ChannelSet` currency.
- `DeviceSetup` — the chosen configuration (mirrors `AudioDeviceSetup`).

Tests (`device_interfaces.cpp`): A/B `ChannelSet` against `juce::BigInteger` (bit-identical across setRange/setBit/count/highestBit), and drive a stub device+type+callback to prove the interfaces are complete + implementable.

## Phase 2a — `dusk::audio::DeviceManager` seam over juce
Pure-dusk API; the impl delegates to a wrapped `juce::AudioDeviceManager` and adapts JUCE's device/type/callback types behind it:
- `JuceDeviceAdapter` / `JuceDeviceTypeAdapter` wrap the juce device + type for query/enumeration.
- `CallbackBridge` forwards the RT audio callback to a dusk `IODeviceCallback` with a context translation, no allocation.
- `ChannelSet <-> juce::BigInteger` and `DeviceSetup <-> AudioDeviceSetup` bridges.
- Backend registration (PipeWire/ALSA on Linux, WASAPI/ASIO/DirectSound on Windows) moved into `initialise()`.
- `juceManager()` is a temporary escape hatch: the MIDI input layer still drives MIDI device enable/disable through the same `juce::AudioDeviceManager`, so it stays until a native MIDI backend lands and `juce_audio_devices` unlinks.

Includes review fixes: stable type-adapter storage keyed by the juce type pointer (a returned `IODeviceType*` no longer dangles across calls), and an idempotent `registerBackends()` guard.

## Verification
- gate 193 → 195 (the two seam files; interface headers are JUCE-free)
- app build 0 new warnings, ctest 389/389, Xvfb selftest 15/15

[...]

PR #90, opened 2026-07-16
De-JUCE device layer Phase 2b: flip AudioEngine onto dusk::audio::DeviceManager

marc-korte:
Phase 2b of replacing `juce::AudioDeviceManager`. AudioEngine now talks to the `dusk::audio::DeviceManager` seam (merged in #89) instead of juce directly. **Behavior-preserving** — the seam wraps juce, so the whole app opens + streams a real device exactly as before.

## What changed
- AudioEngine implements `dusk::audio::IODeviceCallback` (was `juce::AudioIODeviceCallback`) and owns a `dusk::audio::DeviceManager` (was `juce::AudioDeviceManager`).
- Callbacks: `audioDeviceIOCallbackWithContext` → `audioDeviceIOCallback` (dusk `CallbackContext`); `aboutToStart`/`stopped` take dusk `IODevice`; `audioDeviceError` takes `std::string` (added to the dusk interface + bridged).
- Backend registration moved out of AudioEngine into the seam's `initialise()`; saved device state read/restored as an opaque blob.
- ALSA busy-device recovery + hot-unplug detector re-expressed against the dusk API (`getSetup`/`setSetup`/`getCurrentDevice`/`getStateBlob`, `setChangeCallback` replacing the juce `ChangeListener`); `callAsync` → `dusk::callAsync`.
- Offline render + self-test detach/reattach via new `AudioEngine::detach/reattachAudioCallback()` instead of poking the manager with `&engine`. This removed the now-dead `juce::AudioDeviceManager&` members threaded through BounceEngine and the Bounce/Freeze dialogs (net −56 lines).

`getDeviceManager()` returns the seam's `juceManager()` escape hatch, so the selector UI + MIDI input layer (not yet flipped) compile unchanged — Phase 2c moves those, Phase 3 swaps the backing to native and unlinks `juce_audio_devices` on Linux.

## Review
One reviewer-confirmed regression fixed (061f9ba): the startup "device unavailable" message read the device name from `getSetup()` after `initialise()`, which returns the fallback device on failure, so the alert never fired — now reads the intended device from the saved blob.

## Verification
- gate stays 195 (behavior-preserving; module-drop is Phase 3)
- app build 0 new warnings, ctest 389/389
- Xvfb selftest 15/15 with a real UMC1820 opening + streaming, Parallel-Matches-Serial bit-identical

**Owed before merge:** hardware bench on the failure paths (hot-unplug mid-play, xrun recovery, device/SR/buffer swap in settings, ASIO re-pick) — headless can't exercise those.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio-device recovery when the selected device is unavailable, including more reliable fallback behavior.
  * Improved handling of device changes and hot-unplug events to reduce false warnings during normal reconfiguration.
  * Improved audio callback management during saving, exporting, freezing, and DSP setting changes.
* **Improvements**
  * Audio device settings are now restored more reliably between launches.
  * Bounce and freeze workflows now manage audio devices more consistently during rendering.

### DIFF
--- a/src/DuskStudioApp.cpp
+++ b/src/DuskStudioApp.cpp
@@ -486,7 +486,7 @@ static void runHeadlessPipelineTest (const juce::String& pluginPath)
     // re-prepares), so detach FIRST - before prepareForSelfTest mutates DSP
     // state - so the device thread can't process the engine mid-re-prepare
     // (matches runHeadlessSessionPerf).
-    engine->getDeviceManager().removeAudioCallback (engine.get());
+    engine->detachAudioCallback();
     // Don't depend on a real device coming up - prepare directly.
     engine->prepareForSelfTest (sampleRate, blockSize);
 
@@ -655,7 +655,7 @@ static void runHeadlessPipelineTest (const juce::String& pluginPath)
     for (int c = 0; c < numOutChannels; ++c)
         outputPtrs[(size_t) c] = outputs[(size_t) c].data();
 
-    juce::AudioIODeviceCallbackContext ctx {};
+    duskstudio::device::CallbackContext ctx {};
 
     // Two probes:
     //   - Master output peak/RMS - what the device would hear.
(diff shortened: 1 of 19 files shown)


marc-korte:
@coderabbitai review

Issue #143, opened 2026-07-31
DUSKSTUDIO_REPLACE_TEST races the live device callback


marc-korte:
src/DuskStudioApp.cpp:1666 (release/0.12)

The replace test is the only headless gate that never calls removeAudioCallback. The AudioEngine ctor does addAudioCallback(this), so prepareForSelfTest mutates DSP state and the 200 manual audioDeviceIOCallbackWithContext calls at :1720 run concurrently with the live device thread invoking the same callback. runHeadlessPipelineTest and runHeadlessSessionPerf both detach first, with a comment naming this exact hazard.

Fix: detach the callback like the sibling gates.

marc-korte:
Fixed in #156: DUSKSTUDIO_REPLACE_TEST detaches the device callback before driving the engine manually, matching the other headless gates, and exits nonzero on load or swap failure. Ported to main in #158.

Issue #142, opened 2026-07-31
CI: native LV2 silently absent, no release branch coverage, self-test gates cannot fail


marc-korte:
.github/workflows (release/0.12)

1. No Linux workflow installs liblilv-dev, libsuil-dev, lv2-dev (linux-build.yml:42, linux-release.yml:89, linux-sanitizer.yml:58, raspberry-pi-build.yml:35). pkg_check_modules fails, DUSKSTUDIO_NATIVE_LV2 is OFF in every CI job and every shipped Linux tarball, and src/engine/lv2 plus the lv2 tests are never compiled anywhere in CI. The native LV2 host is silently missing from releases behind a bare status message.
2. No workflow carries a release/** branch filter, so pushes to release/0.12, the live branch, get zero build or test. TSan is main-only.
3. The headless self-test gates exit 0 even on FAIL (DuskStudioApp.cpp:1469, 1476, 1483, 1506 never call setApplicationReturnValue), so the CI self-test steps at linux-build.yml:114 and raspberry-pi-build.yml:109 are green by construction. Neither greps for [FAIL].

Fix: install the LV2 deps and fail the build if the LV2 host does not configure, add release/** to the branch filters, and make the headless gates return a real exit code.

marc-korte:
Fixed in #156: every Linux workflow installs lilv/suil/lv2 and configures with DUSKSTUDIO_REQUIRE_NATIVE_LV2=ON (hard configure failure when missing), release/** branch filters run CI on the release branch, and the headless gates set real exit codes. Ported to main in #158.

Issue #138, opened 2026-07-31
Freeze modal hosted on wrong component, hidden mid-render by stage switch


marc-korte:
src/ui/ChannelStripComponent.cpp:3043, related :1859 and :1440 (release/0.12)

freezeModal and hardwareInsertModal are hosted on findParentComponentOfClass, which resolves to ConsoleView, not getTopLevelComponent() like every sibling modal. The dim overlay only covers the console, so transport and menu stay clickable. For the freeze modal, shown with dismissOnClickOutside=false and dismissOnEscape=false, a stage switch (Cmd+3/Cmd+4) hides ConsoleView while the render keeps going and the only teardown controls are now off screen.

Also :1440, the dtor calls closeAndDeleteBodyNow() on the other modals but not hardwareInsertModal or freezeModal, so FreezeDialog's dtor, which cancels a BounceEngine render against AudioEngine, is deferred to a message loop tick and can run after the engine is gone.

Fix: host both on getTopLevelComponent(), add both to the dtor teardown, and close the freeze modal (cancelling the render) on stage switch or make it survive console hiding.

marc-korte:
Fixed in #156: the freeze and hardware-insert modals host on the top-level component, and the strip destructor tears both down synchronously. Ported to main in #158.

Issue #137, opened 2026-07-31
EmbeddedModal::close on a dead host leaves a dangling modal stack entry


marc-korte:
src/ui/EmbeddedModal.h:309 (release/0.12)

close() early-returns when host is null, before removeFromModalStack(), stopListeningForOutsideClicks() and restoreHiddenPluginEditors(). host is a SafePointer, so it nulls when the host component dies with the modal open. The dtor calls close(), so a destroyed modal can leave a dangling pointer in activeModalStack(), which the next close() dereferences (stack.back()->getBody()) and the dim-click guards compare against. The same path leaks a dangling Desktop global mouse listener for the useOverlay=false popups and leaves plugin editors permanently hidden with an orphaned dusk_modalHideCount property.

Fix: run the stack/listener/editor cleanup before the host null check.

marc-korte:
Fixed in #156: close() runs the registry, listener, and hidden-editor cleanup even when the host component died with the modal open, and outside clicks are marked consumed so one click cannot dismiss two stacked layers. Ported to main in #158.

PR #156, opened 2026-07-31
0.12.7 critical wave, batch 4: UI, app shell, CI

marc-korte:
Final batch of the criticals-first wave. Targets release/0.12 for 0.12.7.

## Fixes

- EmbeddedModal: close() on a dead host (a SafePointer that nulls when the host component is destroyed with the modal open) bailed before removing itself from activeModalStack - which the next close() dereferences - and before dropping the Desktop mouse listener and restoring hidden plugin editors. The cleanup runs first; a dead host drops owned objects and callbacks in place with no spurious focus grab. Refs #137.
- Freeze and HW-insert modals: both hosted on the immediate parent (ConsoleView) unlike every sibling, so the dim covered only the console and a stage switch hid the non-dismissable freeze dialog mid-render with its only Cancel/Close controls. Both now host on the top level and join the destructor's closeAndDeleteBodyNow list (FreezeDialog's destructor cancels a render against the engine and must not run on a deferred tick). Bench-verified live on this build. Refs #138.
- DUSKSTUDIO_REPLACE_TEST: the only headless gate that never detached the engine as the device callback, so its manual callback drives raced the live device thread. It detaches first like the pipeline/perf gates and returns 1 on any FAIL. Refs #143.
- Honest headless exit codes: the selftest, instrument, pipeline, and perf gates returned 0 regardless of outcome, making their CI steps green by construction. Each now returns pass/fail through the app exit code (selftest scans its report for FAIL lines; the instrument gate fails on silence; the pipeline gate on a silent strip or master). Refs #142.
- CI: release/** added to every filtered workflow, so pushes and PRs to the live release branch finally build and test (windows-tests' dead develop entry dropped); liblilv/libsuil/lv2 dev packages installed in all four Linux jobs; and DUSKSTUDIO_REQUIRE_NATIVE_LV2=ON in the CI/release configures fails the configure when the deps are missing instead of silently shipping tarballs without the native LV2 host. Refs #142.

## Verification

- App build clean; ctest 386/386.
- Headless self-test under Xvfb exits 0 with zero FAIL lines (the exit code is now load-bearing).
- HW-insert dialog bench-verified on the fixed build.

This PR gives release/** CI to itself and to any still-open wave PRs once merged.

With this and #155 merged, all 18 criticals from the 2026-07-31 scan are dispatched (#126-#143; the two deferred halves - native MIDI input rearchitecture and the RecordManager destructor wait - are documented on their issues).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hardware-insert and freeze dialogs so they remain visible and interactive when switching stages.
  * Fixed modal cleanup to prevent orphaned dialogs and restore plugin editors reliably.
  * Improved self-tests and headless operations to report failures accurately through application exit status.

* **Reliability**
  * Strengthened platform build and test validation across Linux, macOS, Windows, and Raspberry Pi.
  * Improved handling and reporting of missing audio-plugin dependencies during configuration.


marc-korte:
@coderabbitai review

Issue #146, opened 2026-07-31
DSP: tuner and HW insert ping blow the RT budget; assorted meter and latency nits


marc-korte:
src/dsp (release/0.12)

- PitchDetector.h:77 the YIN scan is O((maxLag-minLag) x frameLen) per pushBlock, about 1M inner iterations at the default prepare, independent of numSamples. The early break only fires on voiced input, so unvoiced or noisy input above the gate runs the full scan every callback. The header claims ~10 us per block; it is off by about two orders of magnitude and exceeds a 64-sample buffer budget on its own. Runs whenever the tuner targets a track.
- HardwareInsertSlot.cpp:209 the chirp correlation budget is kCorrelationsPerBlock (256) x chirpLen MACs per callback, 1.2M at 48k and 2.5M at 96k, independent of numSamples.
- HardwareInsertSlot.cpp:174 dry delay setDelay without reset() jumps the read head, a click on every latency slider drag.
- LoudnessMeter.cpp:89 integrated LUFS gates over 100 ms blocks instead of BS.1770-4's 400 ms with 75 percent overlap, so LUFS-I disagrees with R128 compliance meters. Momentary and short-term are correct.
- LoudnessMeter.cpp:120 finishBlock makes two full passes over up to 36000 doubles on the audio thread every 100 ms, unbounded by buffer size.
- MasterBus.cpp:193 no numSamples == 0 early return (all siblings have one); master meters flicker to -inf on device transitions.
- MasteringChain.cpp:169 the oversized-block contract is jassert-only; AuxLaneStrip hard-bails on the identical contract.
- BrickwallLimiter.cpp:84 and MasteringDigitalEq.cpp:51 lround(26.5) = 27 while the true 4x round trip is 26.5 base samples, so mastering latency over-reports by one sample and the "whole number" comments are wrong.
- PitchDetector.h:73 uses std::numeric_limits without including <limits>.

marc-korte:
Status of the nine items, checked against main at e275449 plus PR #166.

Done:

- PitchDetector YIN scan: throttled via scanInterval_ (#164).
- PitchDetector missing `<limits>`: included (#164).
- HardwareInsertSlot dry delay setDelay without reset: resets after retarget (#163).
- MasterBus missing numSamples == 0 return: present (#163).
- MasteringChain oversized-block jassert-only: hard bails now (#163).
[...]

PR #166, opened 2026-08-04
Bound the LUFS gating scan and gate over 400 ms windows (#146)

marc-korte:
Closes the remaining `src/dsp` items on #146.

Note on reading the history: the gating scan was first bounded with a loudness histogram (`87c7f92`), and review found a program shape where bucketing is ~10 LU wrong. `2934300` replaces the buckets with an exact scan, so the histogram does not survive into the final state. Reviewing the diff as a whole is more useful than commit by commit.

## Bound the gating scan (`87c7f92`, superseded in part by `2934300`)

`finishBlock` walked the whole block history twice per 100 ms block to apply BS.1770-4's two gates, so the audio callback's cost grew with program length: an hour in, 36000 doubles scanned twice, unbounded by buffer size.

The two gates behave differently, and the split is what the final design keeps:

- The **absolute** gate is a fixed -70 LUFS threshold. A window's verdict never changes, so its sum is incremental and exact, O(1).
- The **relative** gate moves with the program mean, so its window set has to be recomputed against a threshold only known at publish time.

## Apply the exact relative gate (`2934300`)

The first attempt summarised windows into 0.1 LU buckets to bound that second pass. That cannot be made exact: the threshold can fall inside a bucket, and the windows sharing it are then all included or all excluded. On a program whose quiet material lands there - one loud window against ten 0.01 LU below the threshold - that is **10 LU** of error. No bucket edge or width fixes it, verified:

| rule | reviewer's program | mirror (just above the gate) |
|---|---|---|
| include the threshold bucket | -10.0 LU | exact |
| exclude the threshold bucket | exact | +10.0 LU |
| `ceil` instead of `floor` | -10.0 LU | exact |
| 0.01 LU buckets | -10.0 LU | exact |

Equal-valued windows always share a bucket, so narrowing them does not help.

So the windows are retained and the threshold applied to them directly. Only the relative pass scans, and that scan is budgeted against elapsed audio the same way the hardware-insert ping correlator is, completing inside one 100 ms gating period without any callback carrying it. One unamortised pass measures 27 us, which is 16 % of a 16-sample callback at 96 kHz - that measurement is why it is amortised rather than simply left as one pass.

Retaining windows brings back the one-hour ceiling, which freezes the integrated reading rather than dropping the oldest material. Documented in MANUAL (`14c070c`); it was previously undocumented.

## Gate over 400 ms windows (`6adbda2`)

BS.1770-4 gates 400 ms blocks overlapping by 75 %, one per 100 ms step. The gating ran over the 100 ms sub-blocks instead, which discards short quiet dips that a 400 ms window rides over, so LUFS-I disagreed with R128 compliance meters.

No new state: the sub-blocks are equal length, so the mean of the momentary ring is exactly the 400 ms window's mean square. The first three sub-blocks produce no gating block. Momentary and short-term are unaffected.

MANUAL already claimed "gated per BS.1770", so the code caught up to the documentation rather than the reverse.

## Correct the limiter's group-delay claim (`11166a8`)

The issue reports a one-sample over-report. It is half a sample: `latency()` returns `23.0 + 3.5 = 26.5` at 4x, which integer PDC cannot express, so there is nothing to fix in the code. The stale part was the comment claiming the group delay is a whole number and compensation exact.

## Tests

Three cases, each verified to fail against the defect it guards:

- integrated equals momentary on a uniform program
- the relative gate excludes quiet material, bracketed so it fails both a regression to 100 ms gating (reads back at `loud`) and a broken relative gate (reads ~3 LU low)
- gating is exact for windows sitting on the threshold: gates the meter's own reported windows independently (momentary **is** the 400 ms gating window) across a sweep of quiet levels crossing the threshold, requiring agreement to 0.01 LU. The sweep asserts it reached a near-threshold case so it cannot pass vacuously; reinstating bucketed comparison fails it by 4 LU.

526/526 pass, no new warnings, JUCE gate unchanged at 179 / 9257.

## Review notes

- **API removal:** `isIntegratedCapped()` is gone. It had no callers in src, tests or MANUAL. The ceiling it reported still exists and is now documented in MANUAL instead.
- **Counter widths:** an earlier review round asked for 64-bit gating counters. `absoluteGateCount` increments only inside the `windowCount < kMaxWindows` guard, so it and every scan counter are structurally bounded at 36000. No overflow is reachable.
- **Absolute gate comparison** stays `>`, matching BS.1770-4 clause 3 (`l_j > the absolute threshold`).

## Not covered here

Two notes from the merged ping work (#165) are unaffected by this branch and remain open on #146: `kCorrelationMacsPerSecond` sitting ~30 % above the rate it replaced at 48 k/512, and the brute-force correlator that FFT correlation would cut ~100x.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -972,6 +972,8 @@ Right of the chain are three loudness readouts:
 
 A streaming-platform preset picker (Spotify, Apple Music, YouTube, Netflix, etc.) colour-codes the integrated LUFS and true-peak readings according to that platform's target. Pressing **Reset integrated** clears the integrated reading so you can re-measure from a known point.
 
+The integrated reading measures up to an hour of program. Past that it holds where it is rather than continuing to absorb material, so for anything longer, reset it and measure the section you actually care about.
+
 ## Exporting the master
 
 **Export master…** renders the mastering chain offline. Pick a delivery preset first — **WAV 24-bit at the session rate** (archive/streaming), **WAV 16-bit 44.1 kHz with TPDF dither** (CD spec), or **MP3 320 kbps** — then a destination (defaults to `master.wav` / `master.mp3` in the session folder). A progress dialog shows the output path and a bar; the render runs as fast as the CPU allows and you can cancel mid-render.
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -972,7 +972,7 @@ Right of the chain are three loudness readouts:
 
 A streaming-platform preset picker (Spotify, Apple Music, YouTube, Netflix, etc.) colour-codes the integrated LUFS and true-peak readings according to that platform's target. Pressing **Reset integrated** clears the integrated reading so you can re-measure from a known point.
 
-The integrated reading measures up to an hour of program. Past that it holds where it is rather than continuing to absorb material, so for anything longer, reset it and measure the section you actually care about.
+The integrated reading measures up to an hour of material loud enough to count. Silence and anything below −70 LUFS is discarded by the standard's gate and does not use up that hour, so leaving the meter running between takes costs you nothing. Past the hour it holds where it is rather than continuing to absorb material, so for anything longer, reset it and measure the section you actually care about.
 
 ## Exporting the master
 
--- a/src/dsp/BrickwallLimiter.cpp
+++ b/src/dsp/BrickwallLimiter.cpp
@@ -46,8 +46,10 @@ void BrickwallLimiter::prepare (double sampleRate, int maxBlockSize, double init
     const int bs = std::max (1, maxBlockSize);
 
     // Linear-phase halfband FIR so the limiter doesn't smear transients with
-    // phase distortion; the group delay is a whole number of base-rate samples
-    // so delay compensation stays exact.
+    // phase distortion. At 4x the round trip is 26.5 base-rate samples, which
+    // integer PDC cannot express: osLatencyBase rounds to 27, so the reported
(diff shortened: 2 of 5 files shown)


marc-korte:
@coderabbitai review

marc-korte:
Both findings looked at. One was already fixed, one is a documentation problem rather than a functional one.

**Boundary bin (LoudnessMeter.cpp:119-123).** Already fixed in 2934300, which landed before this review. That comment points at `gateBinFor`, which no longer exists. Bucketing is gone entirely: windows are retained and the exact threshold is applied to them, with the second pass budgeted across callbacks so it stays off any single callback. Your near-threshold distribution is now a test, and it fails by 4 LU if the bucketed comparison is put back.

**One-hour ceiling.** The observation is correct: `windowCount` only counts windows above the absolute gate, so a sparse program can keep integrating past an hour of wall clock. Not changing that, because capping on elapsed windows is worse for how the meter is used. A session left open between takes would freeze the reading an hour after reset even if only a few minutes of audio ever played. Discarding sub-gate material is what the standard's gate already does, so spending capacity on it would buy nothing.

What was wrong is that neither the header nor the manual said which of the two it is. Fixed in 271a22b, plus a test pinning it: trailing silence of any length leaves the reading exactly where it was.

Separately, the meter was cross-checked against ffmpeg's ebur128 on the same 18 s program: integrated -8.99 LUFS vs -8.9, true peak -6.02 vs -6.0. Inside the +/-0.1 LU tolerance in EBU Tech 3341.

PR #169, opened 2026-08-05
Fix UI modal, CLAP editor, timer, and separator bugs

### LINKED ISSUE #149: UI misc: modal teardown gaps, CLAP editor embed gaps, timer dtor gaps
src/ui (release/0.12)

- MainComponent.cpp:997 the shutdown closeAndDeleteBodyNow block covers 10 modals but omits dpImportProgressModal, whose body holds a DpImportJob reference declared before it (destructs first) and a 20 Hz timer.
- ClapPluginEditorComponent.cpp:82 tryEmbed has no re-entry guard; LV2 and VST3 both carry one against the second-UI-instance black-rectangle bug. A synchronous request_resize from set_parent/show re-enters editor.embed().
- ClapPluginEditorComponent.cpp:200 timerCallback has no poll-embed fallback branch; LV2 and VST3 both have one for components created while their lane is hidden (session restore before the AUX tab is shown). CLAP editors in that state never embed.
- MainComponent.cpp:3276 the post-load aux pre-warm checks only CLAP and JUCE slots; LV2 or VST3-only aux sessions skip ensureAuxView and eat the construction stall on the first Aux switch, the exact stall the pre-warm exists to remove.
- DuskContextMenu.cpp:115 the separator uses parent-relative getX() inside local-space paint(), so separators render shifted or clipped. DuskComboBox does it correctly.
- MiniTimelineStrip.cpp:32, CompMeterStrip.cpp:114, AnalogVuMeter.cpp:127, TapeStrip.cpp:255 are Timer subclasses without dtor stopTimer, against the stated house convention.
- TapeMachineModalEditor.h:73 starts a 20 Hz timer in the ctor with no dtor at all; only the base Timer dtor stops it, after the members timerCallback touches are gone, and the owned third-party editor dtor can pump the message loop in that window.

marc-korte:
## Summary

- tear down the DP import progress modal during shutdown and stop six timer-backed components before their own members are destroyed
- make CLAP editor embedding retry safely when visibility changes, with a re-entry guard around the native embed path
- paint context-menu separators in component-local coordinates

## Why

These gaps could leave modal or timer-owned state alive past its intended lifetime, prevent a CLAP editor from embedding after an ancestor became visible, and draw separators at the wrong horizontal position.

## Validation

- application build passed before publication
- full ctest suite passed before publication
- tools/juce-gate.sh passes with no count increases or allowlist additions

Fixes #149

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CLAP editor loading to prevent duplicate windows and support embedding after the surrounding interface becomes visible.
  - Corrected context-menu separator positioning in panel-based layouts.
  - Improved shutdown reliability by ensuring background updates stop cleanly as dialogs, meters, timelines, and other interface components close.
  - Reduced the risk of crashes or unexpected behavior during application exit and interface teardown.

### DIFF
--- a/src/ui/AnalogVuMeter.cpp
+++ b/src/ui/AnalogVuMeter.cpp
@@ -125,7 +125,7 @@ AnalogVuMeter::AnalogVuMeter (const std::atomic<float>* l, const std::atomic<flo
     startTimerHz ((int) kRefreshHz);
 }
 
-AnalogVuMeter::~AnalogVuMeter() = default;
+AnalogVuMeter::~AnalogVuMeter() { stopTimer(); }
 
 void AnalogVuMeter::setReferenceLevelDb (float refDb)
 {
--- a/src/ui/ClapPluginEditorComponent.cpp
+++ b/src/ui/ClapPluginEditorComponent.cpp
@@ -108,14 +108,23 @@ void ClapPluginEditorComponent::tryEmbed()
     // to set_parent/show into a parent that isn't viewable yet - so no pre-warm: build
     // + map when shown, exactly like the JUCE editor path. The kept-alive remap on a
     // later tab switch keeps re-opens instant.
-    if (! loaded || embedded || ! isShowing()) return;
+    // `embedding` guards re-entry. Unlike LV2, a CLAP request_resize can't reach us
+    // synchronously - it only records atomics, and onResize dispatches later from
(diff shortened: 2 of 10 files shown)


marc-korte:
@coderabbitai review

marc-korte:
@coderabbitai review

marc-korte:
@coderabbitai review

PR #175, opened 2026-08-05
Close MIDI input callback window (#157)

### LINKED ISSUE #157: Test flake: JUCE ALSA MIDI input dispatches into a null callback under live MIDI traffic
tests/midi_device_layer.cpp, "MidiInputClient appends a stable Virtual-Keyboard slot" (release/0.12)

Intermittent SIGSEGV, roughly one in ten local runs, only on a machine with live MIDI traffic (hardware sending clock or active sensing). Backtrace lands on JUCE's AlsaMidiHelpers::SequencerThread delivering an incoming event through MidiInput::Impl::consume into a null function pointer inside the bytestream translator callback chain - the event arrives in the window where the test's MidiInputClient::rebuild has enabled the device but the input callback chain is not fully wired.

CI never reproduces it (no MIDI hardware on runners). Not touched by the 0.12.7 wave: the input path is unchanged; only the output bank got the atomic open-state mirror.

Candidate fixes: enable JUCE MIDI inputs only after the callback is registered in MidiInputClient::rebuild, or patch the JUCE-wayland fork's ALSA MIDI input to null-check the dispatch target. Needs a look at the fork's juce_Midi_linux path.

marc-korte:
## What

- build every MIDI input route and the virtual-keyboard slot before enabling any backend input
- stop backend dispatch across a rebuild and restore it only after the finished collector bank is published
- add a hostile test backend that delivers immediately during enable and verifies dispatch fencing

## Why

The reported release/0.12 backtrace was inside JUCE ALSA MIDI input code. That exact Linux path no longer compiles on main because Linux now uses the native ALSA sequencer backend, so this does not claim to repair that removed implementation. It closes the equivalent window in the current Dusk-owned seam: no input is subscribed before its route exists, and no MIDI thread can demultiplex through a half-built bank.

Live MIDI hardware verification remains a manual follow-up because CI has no active input device.

## Validation

- DuskStudio production target builds
- focused MIDI device layer: 53 assertions in 6 test cases
- full Catch2/CTest suite: 550/550 passed
- de-JUCE gate: 179 files / 9,256 uses

Fixes #157

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MIDI device initialization and rebuilding for more reliable input handling.
  * Prevented MIDI events from being processed while device routes are being refreshed.
  * Ensured inputs are enabled only after routing is fully configured.
  * Improved event delivery so connected MIDI inputs receive notes correctly without affecting the virtual keyboard slot.

### DIFF
--- a/src/engine/midi/MidiDevices.cpp
+++ b/src/engine/midi/MidiDevices.cpp
@@ -51,7 +51,12 @@ std::unique_ptr<IMidiOutputBackend> makeOutputBackend()
 // MidiInputClient
 
 MidiInputClient::MidiInputClient()
-    : backend (makeInputBackend())
+    : MidiInputClient (makeInputBackend())
+{
+}
+
+MidiInputClient::MidiInputClient (std::unique_ptr<IMidiInputBackend> backendIn)
+    : backend (std::move (backendIn))
 {
     backend->setReceiver ([this] (const std::string& identifier,
                                   const std::uint8_t* bytes, int numBytes, double timeMs)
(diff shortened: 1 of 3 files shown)


marc-korte:
@coderabbitai review

PR #209, opened 2026-08-06
ci: fold Linux tests into the build job, cache objects with ccache

### LINKED ISSUE #205: CI: fold Linux tests into the build job, add ccache, run tests on arm64
PR feedback on the Linux side takes ~21 minutes because the Catch2 job cold-builds the entire tree a second time on a fresh runner after `needs: build` already made it wait for the first cold build. Same OS, same toolchain, no object sharing, duplicate checkout/apt/pinned-clones. Meanwhile the arm64 job builds on a native runner but never runs the tests at all.

Three changes:

1. Fold the "Catch2 tests (GCC Release)" job into the Linux build job, macOS-style: one runner configures build/ and build-tests/, builds both, runs ctest. Deletes the duplicate setup and compile; Linux verdict drops from ~21 min to ~12 on a cold cache.
2. Add ccache with actions/cache to the Linux jobs (build+tests merged job, TSan with its own cache key since sanitizer objects differ, arm64). Steady-state PR rebuilds drop to a few minutes.
3. Run the test suite on the arm64 job (native ubuntu-22.04-arm runner, no cross-compile obstacle) for arm coverage of the DSP paths.

Keep the pinned JUCE/donor/DPF revs and the De-JUCE ratchet exactly as they are; release workflows untouched. The merged job changes a required-check name, so branch protection needs the new name swapped in when this merges.

marc-korte:
Closes #205.

The Linux side of a PR took about 21 minutes of pipeline time and, in the sample run I measured, 51 minutes of wall clock: the Catch2 job waited on `needs: build`, then queued for a second runner (24 minutes in that run), then cold-built the entire tree again - same OS, same toolchain, same JUCE/sfizz/donor objects as the build job it was waiting for.

Three changes:

- The Catch2 job folds into the Linux build job, the shape macOS already uses. One runner checks out, installs deps and clones the pinned JUCE/donor/DPF once, builds the app, runs the headless self-test, then configures and runs the test suite. The De-JUCE ratchet job is untouched. The test configure flags are carried over unchanged, including `-DDUSKSTUDIO_REQUIRE_NATIVE_LV2=ON`.
- ccache with actions/cache on the three Linux jobs (merged build+tests, TSan, arm64), each with its own cache key since sanitizer and arm64 objects are not interchangeable. No compiler-launcher flags are passed: CMakeLists.txt already finds ccache and sets `CMAKE_C/CXX_COMPILER_LAUNCHER` unconditionally at top level, which both build directories inherit. `ccache -s` runs at the end of each job so the hit rate is visible.
- The arm64 job now runs the test suite too. It builds on a native runner, so there was never a technical reason to skip it; this is the first arm coverage the suite has had.

Expected Linux verdict: about 12 minutes cold, 4 to 6 warm, with the second-runner queue gone entirely.

Branch protection needs an edit at merge time, and it has to happen BEFORE the merge or this PR is unmergeable: ruleset "main protection" currently requires `Catch2 tests (GCC Release, Ubuntu 22.04)` (job no longer exists) and `Build Dusk Studio (GCC Release, Ubuntu 22.04, amd64)` (renamed). Both come out; `Build + tests (GCC Release, Ubuntu 22.04, amd64)` goes in. The arm64 job's rename needs nothing - it is not a required check.

Two knobs left deliberately unturned: `MainComponent.cpp` uses `__DATE__`/`__TIME__`, so ccache will not cache that one translation unit without `CCACHE_SLOPPINESS: time_macros`, which trades a stale build stamp for the saving; and if a hit rate disappoints, the TSan job's 500M cache is the one to raise first, since instrumented objects are much larger.

Note the suite has never run on aarch64 before. If a float-tolerance assertion fails there it is a real finding rather than a workflow bug, and that job is not a required check.

## Summary by CodeRabbit

* **Tests**
  * Raspberry Pi arm64 builds now run the Catch2 test suite and report test results.
  * Linux and sanitizer workflows consistently report compiler-cache statistics, including after failures.

* **Performance**
  * Added compiler caching to Linux, sanitizer, and Raspberry Pi build workflows to reduce repeated build times.

* **CI Improvements**
  * Updated workflow naming and build configuration for clearer coverage of builds and tests.

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -19,10 +19,17 @@ jobs:
         run: tools/juce-gate.sh
 
   build:
-    name: Build Dusk Studio (GCC Release, Ubuntu 22.04, amd64)
+    name: Build + tests (GCC Release, Ubuntu 22.04, amd64)
     # arm64 / Raspberry Pi build-check lives in raspberry-pi-build.yml so each
     # arch only builds once per push.
     runs-on: ubuntu-22.04
+
+    env:
+      # CMakeLists.txt auto-detects ccache and sets the compiler launchers,
+      # so installing it and persisting this directory across runs is the
+      # whole wiring. build-linux and build-tests share the one cache.
+      CCACHE_DIR: ${{ github.workspace }}/.ccache
+      CCACHE_MAXSIZE: 500M
 
     steps:
       - name: Checkout Dusk Studio
@@ -38,7 +45,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends \
-            build-essential cmake ninja-build pkg-config \
+            build-essential cmake ninja-build pkg-config ccache \
             libasound2-dev libjack-jackd2-dev libmp3lame-dev libsndfile1-dev \
             liblilv-dev libsuil-dev lv2-dev \
             ladspa-sdk \
@@ -47,6 +54,14 @@ jobs:
             libxinerama-dev libxrandr-dev libxrender-dev libxss-dev \
             libwebkit2gtk-4.0-dev libgl1-mesa-dev libglu1-mesa-dev xvfb \
             libwayland-dev libxkbcommon-dev libdecor-0-dev
+
+      - name: Cache ccache objects
+        uses: actions/cache@v6
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ccache-linux-gcc-${{ github.sha }}
+          restore-keys: |
+            ccache-linux-gcc-
 
       - name: Clone JUCE (Dusk-owned Wayland mirror, pinned)
         # CI builds against the SAME Dusk-owned mirror snapshot the release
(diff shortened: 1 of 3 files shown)


marc-korte:
Skipping this one: GitHub's cache scoping already prevents it, so there is nothing for a condition or a split key prefix to protect against.

A cache written by a pull request run is created for the merge ref, and per the [dependency caching reference](https://docs.github.com/en/actions/reference/dependency-caching-reference) it "will have a limited scope and can only be restored by re-runs of the pull request". Runs on main and on release branches cannot read it, and neither can other pull requests. The access direction is one-way: a PR may restore from its base branch and the default branch, which is the intended behaviour here and is a read of trusted data, not a write into it.

So `ccache-linux-arm64-` is a shared key *name*, not a shared namespace. The same applies to `ccache-linux-gcc-` and `ccache-linux-tsan-` in the other two jobs, which have the identical shape.

Worth noting as a second layer even if the scoping did not exist: ccache keys its entries on a hash of the preprocessed source, compiler and flags, so a poisoned entry would have to collide with the hash of the exact translation unit being compiled to be served at all.

Issue #218, opened 2026-08-07
Packaging: assert the install set stays scoped


marc-korte:
#215 scoped the install set down to the Dusk payload with EXCLUDE_FROM_ALL on JUCE's subdirectory. Nothing asserts it stays that way.

Two guards worth adding:

- scripts/package-windows.ps1's install gate asserts presence only (bin/DuskStudio.exe, plugin host, LICENSE files). Add an absence check: fail if the staged prefix contains include/ or lib/cmake/, or cap the file count. A future add_subdirectory landing without EXCLUDE_FROM_ALL would ship silently today.
- tests/CMakeLists.txt pulls Catch2 via FetchContent with no EXCLUDE_FROM_ALL, so Catch2's headers, libs and cmake config sit in the build-tests install set. Nothing cpacks that dir, so it is latent, but it is the same class of leak.

Issue #214, opened 2026-08-06
LICENSES.txt completeness: JUCE-bundled libraries (SheenBidi is Apache-2.0), dynamic deps, and two wrong entries


marc-korte:
Follow-up to #179, which credited the notepad's DPF/ImGui/pugl stack and started shipping the license texts in packages. A completeness sweep over what the shipped binary actually links found more attribution-bearing components with no LICENSES.txt entry. All confirmed by symbol in the built binary unless noted.

**Statically linked through JUCE.** The current JUCE entry covers only JUCE-the-framework; JUCE's own LICENSE.md enumerates these bundled libraries:

- SheenBidi - **Apache-2.0**. Highest priority here: Apache-2.0 requires the license copy AND the NOTICE file accompany the binary, so this is an obligation we are not currently meeting, not just a missing credit.
- HarfBuzz (old MIT; Google, Ebrahim Byagowi, Facebook, Mozilla, Codethink, Nokia, Keith Stribley, SIL)
- libjpeg (IJG license, which requires the "based in part on the work of the Independent JPEG Group" acknowledgement)
- libpng (PNG Reference Library License v2)
- zlib (Jean-loup Gailly and Mark Adler)
- libFLAC (BSD-3-Clause)
- Ogg/Vorbis (BSD-3-Clause, Xiph.Org)

**Direct dynamic dependencies** missing from the system-libraries entry: libdbus-1 (AFL-2.1 / GPL-2+ dual, on the link line via dgl-system-libs), libfreetype (FreeType License, which carries a credit clause), libfontconfig (MIT-style).

**sfizz sub-bundles** absent from its Notes sub-list: dr_libs, WavPack, libaiff, cephes, stb_vorbis. Lower priority - all permissive, and sfizz itself is credited.

**Two pre-existing factual errors in the file:**

1. The JUCE entry says "Dual-licensed - GPL-3.0 (used here) OR JUCE Commercial". Wrong for JUCE 8: the open-source arm is **AGPLv3** (JUCE-wayland/LICENSE.md). GPLv3 section 13 supplies the compatibility bridge and the network clause is inert for a desktop application, so this is a documentation-accuracy defect rather than a violation - but it is the most consequential wrong line in the file.
[...]

PR #222, opened 2026-08-07
fix(packaging): exclude JUCE's install rules from the package

### LINKED ISSUE #215: Packaging: cmake --install ships the entire JUCE source tree and juceaide (97 MB)
`cmake --install` lays down 97 MB / 3084 files, of which the actual Dusk Studio payload is about 16 KB of desktop/appdata/mime/icon plus the binaries:

- bin/JUCE-8.0.12/juceaide - a build tool, shipped in a runtime package
- include/JUCE-8.0.12/modules/** - the entire JUCE source tree, including its vendored harfbuzz, sheenbidi, jpglib, pnglib, zlib, flac and oggvorbis sources
- lib/cmake/JUCE-8.0.12/**

Cause: JUCE's add_subdirectory injects its own install rules into our install set. Reaches deb, rpm, MSI and DMG, since all four go through cpack. The tarball is unaffected because scripts/package-tarball.sh hand-stages, and the tarball is the only artifact linux-release.yml publishes today - so this is currently latent rather than shipped.

Three distinct problems if a package is ever cut:

1. FHS and Debian policy: /usr/bin/JUCE-8.0.12/juceaide is a build tool under /usr/bin in a runtime package.
2. Namespace collision: a package named dusk-studio claiming /usr/include/JUCE-8.0.12/** and /usr/lib/cmake/JUCE-8.0.12/** would conflict with any real JUCE development package.
3. Licensing posture: it silently redistributes the complete JUCE source tree plus seven bundled third-party libraries, which is a materially different distribution than linking JUCE, and none of the LICENSES.txt entries describe it.

Fix is small: scope CPACK_COMPONENTS_ALL to our own component, or add_subdirectory JUCE with EXCLUDE_FROM_ALL.

marc-korte:
Fixes #215.

One functional line: JUCE's add_subdirectory gains EXCLUDE_FROM_ALL, which stops CMake emitting the subdirectory's install script. cmake --install drops from 3089 files / 101 MiB to 9 files / 37 MiB; the 9 survivors are the Dusk payload and match the previous install byte for byte apart from relink build ids. No juceaide, no include/JUCE-8.0.12, no lib/cmake/JUCE-8.0.12 in any package.

Safe because JUCE's directory contributes only INTERFACE libraries and an imported juceaide, so nothing we build leaves the default build. Verified from scratch: clean configure + build (warning count identical), cpack TGZ payload is exactly the 9 files, 632/632 tests, juce-gate green. sfizz and DPF already carried EXCLUDE_FROM_ALL; the generated install script now includes nothing.

Packaging surfaces audited: package-tarball.sh hand-stages and never runs cmake --install; package-windows.ps1's install gate asserts only files from our own rules; macOS DragNDrop and DEB/RPM sit on the same scoped install set.

Merge order: land #214 first. The JUCE source tree this removes was the accidental carrier of third-party license texts in the DMG and MSI; #214 makes LICENSES.txt the deliberate carrier.

Follow-up regression guards (assert the install set stays scoped) are #218.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,7 +260,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     unset(_duskstudio_juce_wayland_dir)
 endif()
 
-add_subdirectory(${DUSKSTUDIO_JUCE_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR}/JUCE)
+# EXCLUDE_FROM_ALL keeps JUCE's own install rules out of our install set;
+# without it every cpack package carries juceaide, the whole module source
+# tree and JUCE's CMake package config. The default build is unaffected: the
+# modules are INTERFACE libraries and juceaide is an imported executable.
+add_subdirectory(${DUSKSTUDIO_JUCE_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR}/JUCE EXCLUDE_FROM_ALL)
 
 # App icon. JUCE bakes the PNG into platform-specific resources at
 # build time (.ico on Windows, .icns on macOS, embedded data block on


marc-korte:
@coderabbitai review

Issue #178, opened 2026-08-05
0.13.0 release mechanics: VERSION gate, CHANGELOG, appdata backfill, stale version sites


marc-korte:
Mechanical release chores that will fail or embarrass a `v0.13.0` tag if skipped. 0.13 is the first release cut from main. Note: no `v0.12.7` tag exists — the 0.12.7 wave merged to release/0.12 but was never versioned or tagged (`VERSION` there still reads 0.12.6); decide whether to tag it or fold its notes into 0.13.0.

**Tag-gate blockers**
- `VERSION` reads `0.12.2`; all four release workflows hard-assert `v$(cat VERSION) == $GITHUB_REF_NAME` (`linux-release.yml:81`, `macos-release.yml:57`, `windows-build.yml:63`, `manual-pdf.yml:58`). Tagging without the bump fails every artifact.
- `CHANGELOG.md` has no `[0.13.0]` section and `scripts/bump-version.sh:47-50` aborts without one. 61 commits since v0.12.6 are unrecorded, including native macOS AU (#154), PipeWire backend (#79), live-input monitoring (#77), MIDI hot-plug (#100), SoundFont preset browser (#96), single-pass stems + realtime bounce (#66), and the notepad (#159/#176).
- `packaging/DuskStudio.appdata.xml:33` — top release entry is 0.12.2; main is missing the 0.12.3-0.12.6 entries that exist on release/0.12 (lines 33-52 there). Backfill before running the bump so 0.13.0 lands on top.

**Bump order (script aborts otherwise)**
1. CHANGELOG section → 2. appdata backfill → 3. clean `../JUCE-wayland` (script hard-fails on a dirty fork) → 4. `scripts/bump-version.sh 0.13.0 "<notes>"` → 5. hand edits below → 6. commit → 7. tag on the commit where VERSION reads 0.13.0.

**Hand-edit sites the script does not touch**
- `src/engine/clap/ClapHost.cpp:19` — hardcoded `host.version = "0.12.0"` reported to every hosted CLAP plugin.
- `README.md:25` says "v0.11.1 (Beta)" (six releases stale); `:57` claims "234 Catch2 tests across 58 files" (actual: ~562 across 138); the `:27-52` status table has no notepad row; `:119` references the removed `plugins-main` checkout.
[...]

PR #233, opened 2026-08-09
chore(release): harden the 0.13.0 release mechanics

marc-korte:
Part of #178 (the remaining items there - maintainer email, the bump itself, the tag - are release-day steps done by hand).

- AppStream release dates aligned to their tags in UTC, the convention bump-version.sh itself stamps with `date -u` (0.11.0 -> 2026-06-22, 0.12.2 -> 2026-07-13; 0.12.3 stays 2026-07-15, which is the tag's UTC date). CHANGELOG's 0.12.2 heading synced to match.
- bump-version.sh: the appdata splice anchor now matches the full anchor comment, first occurrence only. The bare-phrase match was reproducibly corruptible: a release note quoting the phrase became a second anchor on the next bump and nested the new entry inside an old description, producing well-formed XML that passed every existing check. The spliced file is validated with xmllint before it replaces the original (control characters pass entity escaping but are XML-illegal; absent xmllint warns and continues), and VERSION is written last so an aborted bump leaves the tree untouched. The next-steps output now includes dating the changelog.
- The macOS, Windows and manual-PDF publishers get the race-checked create/recheck/upload block linux-release.yml already had, so concurrent tag workflows cannot fail on a 422 from a sibling's create. Known trade: a failure between create and upload leaves a published, asset-less release (previously create owned the assets atomically); net better for the race it closes. The single-shot recheck has no retry, matching the existing Linux block. The dead asset-glob length guard in the macOS/Windows copies (always nonzero because the SHA256SUMS literal is in the array) diverges from Linux's validated-glob version; left as is, both platforms catch empty asset sets earlier.
- README test counts corrected (660 cases across 143 files, counted).
- New release-mechanics contract test: exercises the real bump-version.sh against a fixture (splice placement above existing entries, decoy anchor inside an old description, VERSION written, control-character note aborts leaving VERSION and appdata untouched) and pins the four publishers' create/recheck/upload shape. It does not audit the wider release inventory (appdata ordering across files, changelog agreement); scope is the script and the publisher blocks.

One wrinkle for release day: the script stamps the appdata with the bump day's UTC date; tagging on a later day recreates the date drift this PR cleans up. Bump and tag on the same day, or expect a one-day skew.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved release publishing reliability by handling concurrent release creation and ensuring installers, PDFs, and checksums are uploaded correctly.
  - Prevented invalid release metadata from being saved during version updates.

- **Documentation**
  - Updated release dates and application metadata for versions 0.12.2 and 0.11.0.
  - Refreshed README test-count and repository information.

- **Tests**
  - Added automated validation for release workflows, version updates, changelog entries, and metadata integrity.

### DIFF
--- a/.github/workflows/macos-release.yml
+++ b/.github/workflows/macos-release.yml
@@ -217,16 +217,19 @@ jobs:
             exit 1
           fi
 
-          # Create the release on first push of this tag, else add the
-          # assets (idempotent re-run). --clobber overwrites a same-named
-          # asset from a previous attempt.
           if ! gh release view "$TAG" --repo "$RELEASES_REPO" >/dev/null 2>&1; then
-            gh release create "$TAG" \
+            if ! create_err=$(gh release create "$TAG" \
               --repo "$RELEASES_REPO" \
               --title "Dusk Studio $TAG" \
-              --notes "$NOTES" \
-              "${ASSETS[@]}"
-          else
-            gh release upload "$TAG" --repo "$RELEASES_REPO" --clobber "${ASSETS[@]}"
+              --notes "$NOTES" 2>&1); then
+              if gh release view "$TAG" --repo "$RELEASES_REPO" >/dev/null 2>&1; then
+                echo "release create raced with another release job; the release now exists, uploading."
+              else
+                echo "::error::gh release create failed for ${TAG} and no release exists:" >&2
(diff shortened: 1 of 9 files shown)


marc-korte:
@coderabbitai review

Issue #231, opened 2026-08-08
Stale-reap editor teardown still releases the container with the plugin view inside it (VST3 + AU)


marc-korte:
Follow-up from the #195 review. On the stale-owner reap path (instance already disposed), abandonPlugin() nulls the plugin-side handle but close() then runs removeFromSuperview + release on the host-owned container. The plugin's view is still a subview at that moment, so AppKit propagates viewWillMoveToWindow:/viewDidMoveToWindow down the detached subtree - and AUv2 Cocoa views commonly run AUListenerDispose / AudioUnitRemovePropertyListener from exactly those hooks, i.e. plugin code executing against a disposed instance. Refcount safety holds (the leaked open() retain keeps dealloc from running); the residual is the window-detach notification surface only.

This is inherited from the shipped #181 contract - Vst3Editor::close() has the same shape on macOS - so it must be fixed for VST3 and AU together, not one format alone. Candidate: abandonPlugin() also relinquishes the container (nil without release on the stale path, accepting the container leak), so close() no-ops after abandon; or detach the plugin view from the container inside abandonPlugin() without messaging beyond removeFromSuperview, and measure whether that is actually safer.

Not reachable on Linux. Needs a macOS pass to decide severity - the live-path (session load) ordering is benign because the unit is still alive when the detach notifications fire.

Issue #181, opened 2026-08-05
Native plugin editor lifetime: use-after-free on instance swap (CLAP/LV2/VST3)


marc-korte:
The native plugin editor components hold raw instance pointers with no owner or generation check, while several non-UI paths destroy the instance underneath them. One format already has the fix shape; the other three do not.

**Use-after-free on instance swap (critical)**
- `src/ui/ClapPluginEditorComponent.cpp:246-249` caches raw `plugin`/`hostPtr` into `ClapInstance::hostObj`. The instance can be destroyed by: undo replay (`src/session/RegionEditActions.cpp:925`), session restore (`src/engine/AudioEngine.cpp:1992/2030/2063/2095/2114/2137/2146`), and one-host-per-insert eviction (`src/dsp/ChannelStrip.cpp:546/585/620/658/694`).
- The 60 Hz `editor.pump()` then ticks a retired pointer; the `else { tryEmbed(); }` added in #169 additionally calls `gui->set_parent` on it every tick, and `~ClapPluginEditorComponent` writes through it (`hostPtr->setCallbacks(nullptr)`).
- The repo already solved this for the multisample editor: `ChannelStripComponent::syncMultisampleEditorOwner()` (`src/ui/ChannelStripComponent.cpp:2132`, `:3031`). There is no CLAP/LV2/VST3 equivalent. Same shape exists in `src/ui/Lv2PluginEditorComponent.h:53` and `Vst3PluginEditorComponent`.

**LV2 leak-on-close graph too small (high)**
- `src/engine/lv2/Lv2Editor.cpp:296-319` leaks only the two suil handles, then destroys `impl` — but the leaked UI holds `Impl*` as both its `SuilController` and `ui:resize` handle (`:47-49`, `:63-65`, `:71-84`), and `uiResize` dereferences `self->owner` unguarded (`:82`). The macOS TU does it correctly (`Lv2Editor_Mac.mm:274-289`: leaks the whole `Impl` and nulls `owner` first). Walk the reachable pointer graph and leak exactly that set.

**Teardown under a live leaked GUI (medium)**
[...]

PR #234, opened 2026-08-09
fix(au): give the AU editor the native-owner lifetime contract

### LINKED ISSUE #195: AU editor has the same instance-swap lifetime hole fixed for CLAP/LV2/VST3 in #181
Follow-up to #181: the AU editor path (`AuPluginEditorComponent` / `AuEditor`, macOS-only) has the identical lifetime shape — a cached raw instance pointer with no owner/generation check while undo replay, session restore, and insert eviction can destroy the instance underneath the editor poll.

The fix for #181 (branch fix/181-native-editor-lifetime) introduced the shared mechanism: `NativeInsertSlot::generation()` + `src/ui/NativeEditorOwner.h` stamp + `syncNativeEditorOwners()` reap. Applying it to AU is mechanical — stamp on open, reap in the same polls, plus an `AuEditor::abandonPlugin()` that nulls the plugin-side handles without calling through them — but it is macOS-only code that cannot be compile-verified on the Linux box, so it was deliberately left out of #181's branch rather than shipped blind.

Do this on macOS (or at minimum let macOS CI compile it and Marc smoke-test an AU editor open/close across a session reload).

**Constraint:** zero new JUCE — use the `dusk::` seams; any unavoidable allowlist count change is a hand edit justified in the PR body.

marc-korte:
Closes #195.

Ports the #181 stamp/reap mechanism to the AU editor, which still polled a cached instance pointer while session load, undo replay and insert eviction could destroy the Audio Unit under it. The editor stamps {slot, generation} at attach, checks staleness on the pump tick, the embed, the bounds push and the visibility flip, and is reaped in the same syncNativeEditorOwners pass as CLAP/LV2/VST3. NativeAuSlot needs no change: every AU mutation routes through the base slot's load/unload/leakForShutdown, all of which bump the generation (verified across ChannelStrip, AuxLaneStrip, RegionEditActions and AudioEngine; reactivate keeps the same AudioComponentInstance, so the stamp stays fresh).

AuEditor::abandonPlugin() nulls the plugin-side view without messaging it; the retain taken at open() is deliberately leaked so the view's dealloc, where AU views tear down their listeners against the unit, cannot run against a disposed instance. This is refcount-for-refcount the VST3 contract (Vst3Editor::abandonPlugin drops the IPlugView via take() without Release). close() still detaches and releases the host-owned container on every path. The visibility guard is AU-only by design: unhiding the container draws the plugin's Cocoa subview synchronously, unlike the VST3/X11 case where reveal touches only a host-owned foreign window.

Verification and its limits, stated plainly:
- Linux: app build clean, 649/649 ctest, juce-gate 179 files / 9239 uses unchanged, selftest 15/15 ALSA under Xvfb. However, every AU hunk is behind the Apple-only gate, so zero lines of the new AU code compiled in any of that. macOS CI on this PR is the first compile.
- Owed before this ships in a release: an AU editor smoke on real macOS - open an editor, reload the session, replace and evict the insert, undo across the replacement, quit with the editor open.

Residual, tracked as #231: on the stale-reap path the container release still propagates window-detach notifications through the plugin's view. Inherited from the shipped #181 contract (VST3 has the same shape), so it is filed to fix for both formats together rather than diverging AU here.

### DIFF
--- a/src/engine/au/AuEditor.h
+++ b/src/engine/au/AuEditor.h
@@ -25,6 +25,7 @@ public:
     void setBounds (int x, int y, int width, int height);
     void reveal();
     void hide();
+    void abandonPlugin() noexcept;
     void close();
     void pump();
 
--- a/src/engine/au/AuEditor.mm
+++ b/src/engine/au/AuEditor.mm
@@ -195,6 +195,18 @@ void AuEditor::pump()
     if (onResize) onResize (width, height);
 }
 
+void AuEditor::abandonPlugin() noexcept
+{
+    // The Audio Unit behind this view may already be disposed. Leaking the
+    // retain taken in open() guarantees the view's -dealloc - where AU views
+    // run their listener teardown against the unit - never executes; close()
+    // still detaches and releases the container this host owns, which does
+    // message the detached subtree, but only through plain NSView hierarchy
+    // notifications.
+    impl->pluginView = nil;
+    impl->embedded = false;
+}
+
 void AuEditor::close()
 {
     if (impl->pluginView != nil)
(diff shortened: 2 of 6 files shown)


marc-korte:
@coderabbitai review

PR #248, opened 2026-08-10
perf: reduce hardware insert ping callback cost

### LINKED ISSUE #167: DSP: hardware insert ping correlator cost and search algorithm
Two things came out of the hardware insert ping work in #165. Neither is a defect, both are worth a second look.

## The budget constant is higher than what it replaced

`HardwareInsertSlot::kCorrelationMacsPerSecond` is 1.5e8. The fixed 256-lags-per-block count it replaced worked out to about 1.15e8 MACs/s at the default 48 kHz / 512, so the most common configuration got roughly 30 percent more expensive, while the point of #165 was to bound the cost rather than raise it.

Measured the correlator loop at 9.5e8 MACs/s on this machine (serial dependent FMA chain, no vectorisation, since a float reduction will not vectorise without fast math). So 1.5e8 is about 16 percent of one core here, and closer to 40 percent on the Pi, sustained for the length of a ping.

Lowering it to 1.15e8 restores parity with the old default-config cost. The tradeoff is ping duration: at 96 kHz / 64 that would go from about 1.3 s to about 1.8 s, with no progress feedback beyond the button text. Worth deciding deliberately rather than leaving at a number nobody picked.

Note that `tests/hardware_insert.cpp` pins block counts for the 96 kHz / 64 case, so changing the constant means retuning those bounds.

## The correlator is brute force

The lag search is 16385 candidates times 4800 taps, about 7.9e7 MACs per ping at 48 kHz, and 1.57e8 at 96 kHz where the chirp is 9600 samples.

FFT-based correlation is O(N log N) and would cut that by roughly 100x, which makes the budget constant above irrelevant. A coarse-to-fine search is the cheaper option but needs care: decimating a 20 Hz to 20 kHz sweep aliases badly, so it wants a lowpass first, and the correlation peak gets less sharp.

This is real DSP work rather than a tweak, and the current version is correct and bounded, so there is no urgency.

marc-korte:
## Summary
- reduce the hardware-insert ping correlation budget from 1.50e8 to 1.15e8 MAC/s, restoring the former 48 kHz / 512 callback cost
- preserve the exhaustive scalar search over all 16,385 inclusive lag candidates and the existing detection semantics
- pin exact callback pacing and add maximum-latency, inverted-polarity, and low-return threshold coverage
- update documented ping timing for the lower scheduled callback workload

The scheduled correlation workload is 23.33% lower. Measured pacing is approximately 1.141s at 48 kHz / 512 and 1.639s at 96 kHz / 64.

## Validation
- focused hardware-insert suite: 1,071 assertions / 9 cases passed
- production and test builds
- serial CTest suite: 652/652 passed in 28.03s
- private-Xvfb application self-test: exit 0
- performance guard: no material regression; scalar correlation kernel remains approximately 164ms
- de-JUCE gate: 179 files / 9239 uses, no new literal `juce::`
- `git diff --check`

## Manual follow-up
- physical interface/external-hardware loopback was not available for this pass and remains owed; simulated loopback coverage is green across the named timing and signal boundaries

Closes #167

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hardware-insert ping processing to provide more accurate latency detection across supported sample rates and block sizes.
  * Improved handling of returned signals, including polarity changes and low-level returns, for more reliable hardware-insert measurements.

* **Documentation**
  * Updated estimated measurement durations to approximately 1.1 seconds at 48 kHz and 1.6 seconds at 96 kHz with 64-sample blocks.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -1623,7 +1623,7 @@ The status label shows one of:
 - `No input signal - check routing / device.` — the selected return input is unavailable or invalid, so capture cannot run.
 - `Ping failed - check level / cables.` — the return input is valid, but the capture was silent, too quiet, or contained no clear correlation peak.
 
-Measurement time varies with sample rate and block size; the tested range is roughly 1.0 seconds at 48 kHz and 1.3 seconds at 96 kHz with 64-sample blocks. Correlation is spread across audio callbacks so it stays within its processing budget; smaller buffers use more callbacks, with fractional work carried between them.
+Measurement time varies with sample rate and block size; the tested range is roughly 1.1 seconds at 48 kHz and 1.6 seconds at 96 kHz with 64-sample blocks. Correlation is spread across audio callbacks so it stays within its processing budget; smaller buffers use more callbacks, with fractional work carried between them.
 
 Run the ping after any change to your interface routing or your external gear's settings. There is no automatic re-ping on session load; you ping when you set up the insert and re-ping if anything in the chain changes.
 
--- a/src/dsp/HardwareInsertSlot.h
+++ b/src/dsp/HardwareInsertSlot.h
@@ -70,9 +70,10 @@ public:
     // chirpLength + kMaxDelaySamples, so the correlator's lag search
     // covers the full latency-slider range at any sample rate.
     static constexpr int kChirpMaxSamples   = 9600;
-    // Correlation budget in multiply-accumulates per second of audio. Credit
-    // follows callback duration, and each candidate lag consumes chirpLength MACs.
-    static constexpr double kCorrelationMacsPerSecond = 1.5e8;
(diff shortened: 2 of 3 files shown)


marc-korte:
Review follow-up:

- The exact `kCorrelationMacsPerSecond == 1.15e8` assertion is intentional: this test pins the public callback-budget contract, and 115,000,000 is exactly representable as a double. The suggested `1e-9` absolute tolerance is also smaller than one ULP at this magnitude, so it would not meaningfully relax the assertion.
- The loopback boundary sections remain grouped under the existing exact round-trip test case because they exercise the same simulated-loopback contract and shared harness. Splitting them would be test-organization churn without changing coverage or diagnostics.

No code change is warranted for these two style-only findings.

PR #323, opened 2026-08-19
Link libsodium statically in the Linux release build

### LINKED ISSUE #322: Linux binaries hard-require libsodium.so.23 and will not start on non-Ubuntu distros
## Problem

The shipped Linux binaries hard-require `libsodium.so.23` and do not bundle it. That soname is Ubuntu 22.04's libsodium 1.0.18. Every distro shipping 1.0.19 or newer provides `libsodium.so.26`, so on openSUSE, Fedora and Arch the app does not start — it fails in the dynamic loader before `main`:

```
error while loading shared libraries: libsodium.so.23:
cannot open shared object file: No such file or directory
```

Confirmed on both architectures. `readelf -d` on the shipped `DuskStudio` lists `libsodium.so.23` under `NEEDED` in the x86_64 and aarch64 tarballs alike, and neither binary carries an `RPATH` or `RUNPATH`, so there is no path by which a bundled copy could be found even if one were added to the tarball.

This is a hard launch failure, not a degraded feature. It is also invisible to CI: every Linux workflow runs on `ubuntu-22.04`, the one distro where the soname resolves.

## Why Linux only

Both other platforms already link libsodium statically. The comment at `CMakeLists.txt:638-640` states the split as a deliberate design:

> Windows consumes the static vcpkg manifest prefix, Linux uses the distro library, and macOS links a deployment-target-matched static archive so the app has no extra dylib.

- **macOS** — `CMakeLists.txt:644-661` searches only for `libsodium.a` and raises `FATAL_ERROR` if it is absent. `macos-release.yml:84-95` builds libsodium 1.0.22 from pinned, SHA-256-checked source, and the workflow greps `CMakeCache.txt` to assert the static archive was actually selected.
- **Windows** — static vcpkg triplet.
- **Linux** — `libsodium-dev` from apt, resolved by `find_library(NAMES sodium libsodium)` at `CMakeLists.txt:677`, which prefers the `.so`.

So the machinery to fix this already exists and is proven on macOS. Linux is the only platform that did not adopt it.

libsodium has exactly one consumer: `src/engine/sfz/SfzCatalogEnvelope.cpp`, which verifies detached Ed25519 signatures on the SFZ catalog.

## Fix

Mirror the macOS arrangement for the Linux *release* builds, without forcing it on people building from source.

1. Let the static path be selected on Linux. Today the `libsodium.a` lookup lives inside `if(APPLE)`. Extend it so a Linux configure that is given `DUSKSTUDIO_SODIUM_ROOT` prefers the static archive there, and falls back to the existing `find_library` behaviour otherwise.
2. Keep the distro library as the default for an unconfigured source build. A from-source build on a developer's machine, and any future distro packaging, should keep linking the system libsodium — that is correct behaviour for a distro package and should not regress.
3. In `linux-release.yml`, build the same pinned libsodium 1.0.22 from SHA-256-checked source that `macos-release.yml` already builds, for both the x86_64 and aarch64 jobs, and pass `-DDUSKSTUDIO_SODIUM_ROOT`.
4. Assert the result twice, because a silent fallback to the shared library is exactly the failure mode that shipped:
   - grep `CMakeCache.txt` for the static archive, mirroring `macos-release.yml:130-131`;
   - assert `readelf -d` on the packaged binary lists no `libsodium` under `NEEDED` at all. This is the check that would have caught the bug, and it belongs in the release workflow regardless of how the link is arranged.

Rejected alternative: bundling `libsodium.so.23` in the tarball with an `$ORIGIN` RPATH. It works, but it ships a third-party shared library, adds an RPATH to a binary that currently has none, and leaves a pinned 1.0.18 in the payload that must then be tracked for CVEs. Static linking a pinned 1.0.22 is what the other two platforms already do, and it keeps the tarball a single self-contained tree.

Also viable but larger: #286, "Allow builds without signed SFZ catalog support", which would make libsodium optional entirely. That is worth doing on its own merits, but it does not fix the shipped binary — a build *with* catalog support still has to link something.

## Release impact

`v0.13.0` is published but not announced, and its Linux assets have 2 and 3 downloads, essentially all internal verification traffic. The tarballs must be rebuilt and republished once this lands. Whether that is a moved `v0.13.0` tag or a `v0.13.1` is a call to make when the fix is ready.

## Acceptance

- `readelf -d` on both shipped Linux binaries lists no `libsodium` under `NEEDED`.
- The release workflow fails if a Linux release build ever links libsodium dynamically again.
- A plain source build with `libsodium-dev` installed and no `DUSKSTUDIO_SODIUM_ROOT` still configures and links, unchanged.
- The tarball starts on a distro shipping `libsodium.so.26`.

marc-korte:
Closes #322.

The portable Linux tarballs carried a `NEEDED` entry for `libsodium.so.23`, Ubuntu 22.04's soname, and bundled nothing. Any distro on libsodium 1.0.19 or newer provides `libsodium.so.26`, so on openSUSE, Fedora and Arch the binary died in the dynamic loader before `main`. Both architectures, and neither binary carries an `RPATH`, so no bundled copy could have been found either.

No CI job caught it because every Linux workflow runs on `ubuntu-22.04`, the one distro where the soname resolves.

## Approach

Windows and macOS already link a static libsodium; Linux was the only platform that did not. `CMakeLists.txt:638` described that split as deliberate, and for a distro package it is right. It is wrong for a portable tarball.

- `CMakeLists.txt` widens the existing static-archive branch from `if(APPLE)` to `if(APPLE OR (UNIX AND DUSKSTUDIO_SODIUM_ROOT))`. A source build with `libsodium-dev` and no root set is unchanged, which is what distro packaging wants. Windows is untouched twice over: it passes no `DUSKSTUDIO_SODIUM_ROOT`, and `UNIX` is false there.
- `linux-release.yml` builds the same pinned, SHA-256-checked libsodium 1.0.22 that `macos-release.yml` already builds, for both matrix legs, and passes the prefix to the app and test configures.

Rejected: bundling `libsodium.so.23` with an `$ORIGIN` RPATH. It ships a third-party shared object, adds an RPATH to a binary that has none, and leaves a pinned 1.0.18 in the payload to track for CVEs.

The tradeoff worth naming: a static libsodium means a CVE needs a Dusk Studio rebuild rather than arriving through the distro. That burden already exists on macOS and Windows, and it is the lesser cost against a binary that does not start at all.

## The check was backwards

`Verify libsodium runtime linkage` asserted that `libsodium.so.` **was** a `NEEDED` entry, so it passed on precisely the arrangement that shipped broken. It came in with #255 with no stated rationale and reads as a "did it link at all" smoke test from when dynamic was the only Linux path.

It now asserts the absence, across every executable under the artefacts tree rather than one hardcoded path, and the configure step is separately checked for having selected the archive. `libsodium-dev` stays installed on the runner on purpose: the assertion then proves the static archive won with a shared library present, which is a stronger test of the CMake selection than removing it would be.

## Docs corrected in the same pass

- `LICENSES.txt:465` said Linux release builds "link the distro shared object".
- `packaging/README-linux.txt:37` told users to install `libsodium23`. That one ships inside the tarball.

The pinned SHA matches the audit reference already recorded in `LICENSES.txt` (1.0.22, `adbdd8f1...`), so there is no new supply-chain entry.

## Verification

The workflow cannot run outside a tag, so the CMake block was extracted into a fixture and driven directly:

| Check | Result |
|---|---|
| Linux with `DUSKSTUDIO_SODIUM_ROOT` selects `libsodium.a`, and the CI `CMakeCache` grep matches it | pass |
| Linux without a root still resolves the distro `.so`, source builds unchanged | pass |
| Assertion loop against a dynamically linked binary | exits 1 with the error |
| Assertion loop against a statically linked binary | exits 0 |

Not covered locally, and first exercised on the tag run: the libsodium autotools build has never run on `ubuntu-22.04-arm`, and `--with-pic` is new here (macOS does not need it, a Linux PIE does).

## Release impact

`v0.13.0` is published but unannounced, with 2 and 3 downloads on the Linux assets. The tag moves to this commit and the tarballs are rebuilt. The consolidated `SHA256SUMS` must be regenerated afterwards, since both Linux hashes change.

## Summary by CodeRabbit

* **New Features**
  * Linux release packages now include libsodium directly, eliminating the need for a separately installed runtime library.
  * Signed SFZ catalog authentication works without additional host dependencies.
* **Bug Fixes**
  * Release validation now confirms the packaged application is self-contained and does not depend on an external libsodium library.
* **Documentation**
  * Updated Linux requirements to remove the previously required libsodium package and runtime library installation.

### DIFF
--- a/.github/workflows/linux-release.yml
+++ b/.github/workflows/linux-release.yml
@@ -224,6 +224,34 @@ jobs:
             exit 1
           fi
 
+      - name: Build static libsodium (pinned)
+        # The distro package is the right dependency for a distro build, but
+        # its soname is the distro's: jammy carries libsodium.so.23 and any
+        # distro on 1.0.19+ carries .so.26, so a tarball linked against the
+        # apt library fails in the loader everywhere but Ubuntu. Link the
+        # audited source instead, matching what Windows and macOS already do.
+        env:
+          SODIUM_VERSION: 1.0.22
+          SODIUM_SHA256: adbdd8f16149e81ac6078a03aca6fc03b592b89ef7b5ed83841c086191be3349
+        run: |
+          set -euo pipefail
+          archive="${RUNNER_TEMP}/libsodium-${SODIUM_VERSION}.tar.gz"
+          source_dir="${RUNNER_TEMP}/libsodium-source"
+          prefix="${RUNNER_TEMP}/libsodium"
+          curl --fail --location --retry 3 \
+            --output "${archive}" \
+            "https://download.libsodium.org/libsodium/releases/libsodium-${SODIUM_VERSION}.tar.gz"
+          echo "${SODIUM_SHA256}  ${archive}" | sha256sum --check
+          mkdir -p "${source_dir}"
+          tar -xzf "${archive}" -C "${source_dir}" --strip-components=1
+          cd "${source_dir}"
+          # Position-independent, because it is linked into a PIE executable.
+          ./configure \
+            --prefix="${prefix}" --disable-shared --enable-static --with-pic
+          # Keep this small autotools build aligned with the three-core runner.
+          make -j3 check
+          make install
+
       - name: Configure (Release)
         run: |
           cmake -S . -B build-linux -G Ninja \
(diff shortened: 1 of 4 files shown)


marc-korte:
@coderabbitai review

PR #324, opened 2026-08-19
ci: bound every apt call with timeout(1) and switch mirrors on failure

marc-korte:
Run 32272023755 ("Catch2 tests (TSan)") sat in "Install Linux build deps" for 70+ minutes before being cancelled, despite the action advertising "mirror bypass and retries". Two defects:

1. **No `timeout(1)`**: `Acquire::http::Timeout` only bounds socket inactivity. A mirror that dribbles bytes hangs `apt-get` forever, so the retry loop never gets its second attempt. Every apt invocation is now wrapped in `timeout(1)` (120 s update / 300 s install, three attempts), so the worst case is minutes, never the job timeout.
2. **Permanent mirror pinning**: the action pinned `archive.ubuntu.com`; azure and archive take turns being the sick one (both directions observed within one day on the plugins repo). The installer now switches to the other mirror in the machine's family after the first failed attempt: `azure.archive.ubuntu.com <-> archive.ubuntu.com` on x86, `mirrors.mit.edu <-> ports.ubuntu.com` on arm64 (normalized onto MIT primary with IPv4 forced, preserving the raspberry-pi workflow's proven behaviour).

The logic is ported from dusk-audio-plugins' hardened `apt_install.sh` (dusk-audio-plugins#191 plus its fleet-wide follow-up dusk-audio-plugins#196) and lives in `.github/actions/apt-deps/apt_install.sh`; the action is now a thin wrapper. `raspberry-pi-build.yml` drops its own inline untimeouted retry loop and uses the shared action. Also added: dpkg presence verification after installs, so a silent miss fails at the install step with a clear message.

Verified with actionlint (the single remaining finding, `linux-release.yml:282` SC2206, is pre-existing), YAML parse, and `bash -n` on the script. No raw `apt-get` remains in any workflow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package installation reliability during Raspberry Pi builds.
  * Added automatic mirror switching, IPv4 handling, retries, and operation timeouts.
  * Added support for ARM-specific package mirrors and common APT source formats.
  * Improved validation and error reporting for failed or missing package installations.

* **Maintenance**
  * Centralized dependency installation behavior for more consistent build workflows.

### DIFF
--- a/.github/actions/apt-deps/action.yml
+++ b/.github/actions/apt-deps/action.yml
@@ -17,4 +17,13 @@ runs:
   steps:
     - name: apt-get with timeouts, retries, and mirror switching
       shell: bash
-      run: '"${GITHUB_ACTION_PATH}/apt_install.sh" ${{ inputs.packages }}'
+      # The list reaches the shell through the environment, not through
+      # expression interpolation: a value spliced into the run body becomes
+      # shell syntax, so a metacharacter in it would execute. Word splitting
+      # on the unquoted expansion is what turns the list into arguments, and
+      # set -f keeps a package name from globbing against the workspace.
+      env:
+        PACKAGES: ${{ inputs.packages }}
+      run: |
+        set -f
+        "${GITHUB_ACTION_PATH}/apt_install.sh" $PACKAGES
--- a/.github/actions/apt-deps/action.yml
+++ b/.github/actions/apt-deps/action.yml
@@ -1,10 +1,13 @@
 name: Install apt packages
 description: >
-  apt-get update + install with the Azure apt mirror bypassed, bounded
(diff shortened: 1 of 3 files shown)


marc-korte:
Updated after a live failure surfaced two gaps: the hosted runners route apt through the /etc/apt/apt-mirrors.txt mirrorlist (so sources-only rewrites never actually left azure), and a timed-out install can wedge dpkg ("dpkg was interrupted"), which no retry recovers from. The branch now purges azure from every mirror file including the mirrorlist, uses stock Ubuntu repos only (MIT rotation removed per policy; Ubuntu 22.04 is supported until 2027), keeps IPv4 forced on arm64, and runs a bounded dpkg --configure -a before each retry.

Issue #335, opened 2026-08-22
Notepad on Windows is unverified on real GPU hardware


marc-korte:
The notepad is now verified on Windows only under Mesa's llvmpipe, a software rasteriser, in a VM. That proves the code is correct on a conforming OpenGL 4.6 context: the editor renders, accepts typed text, and the event pump keeps running.

It does not prove the notepad works on a machine with a vendor OpenGL driver, because no such machine has run it. The three Windows graphics stacks seen so far:

| Renderer | Result |
| --- | --- |
| GDI Generic 1.1 (bare VM adapter, RDP session) | refuses cleanly, DAW survives (#331) |
| Mesa D3D12, the OpenGL Compatibility Pack | ends the application (#332) |
| Mesa llvmpipe, conforming GL 4.6 | works |

A consumer GPU with an AMD, Intel or NVIDIA driver should behave like the llvmpipe column, but that is an expectation rather than a result.

What is needed is one run on real Windows hardware: install the MSI, open the notepad, type into it, switch to source view and back, close it, and confirm the text lands in `notepad.md`. Worth capturing the GL identity line from stderr as well, which now reports vendor, renderer, version, GLSL version and max texture size.

`/release` Step 10 says to report Windows as untested rather than passed when no machine is available. If 0.13.1 ships before this happens, the release notes should say the notepad is unverified on Windows GPU hardware rather than implying it was checked.

marc-korte:
Still open, and the win11 VM cannot close it. Recording why, so the next attempt does not repeat the work.

The VM reaches exactly two OpenGL renderers, and #337 refuses both by design:

| Renderer | Result with the 0.13.1 MSI |
| --- | --- |
| GDI Generic 1.1 (the VM's default: no ICD registered, QXL exposes no `OpenGLDriverName`, no Mesa present) | refused cleanly, DAW survives |
| Mesa D3D12 via the Compatibility Pack | refused by name, DAW survives (#332) |

So there is no positive control on this machine: nothing proves the notepad *renders* on Windows, only that it declines safely.

The llvmpipe row in the table above cannot be reproduced from the Store pack. `Microsoft.D3DMappingLayers` ships `OpenGLOn12.dll` only, and any Mesa override makes it bail back to GDI Generic. Tried and confirmed useless here:

```
LIBGL_ALWAYS_SOFTWARE=1                                  -> GDI Generic 1.1
GALLIUM_DRIVER=llvmpipe                                  -> GDI Generic 1.1
[...]

PR #337, opened 2026-08-23
Close the 0.13.1 release blockers

### LINKED ISSUE #332: [0.13.1 blocker] Notepad ends Dusk Studio on Mesa D3D12 (OpenGL Compatibility Pack)
Opening the session notepad ends Dusk Studio on a Windows machine whose OpenGL comes from Mesa's D3D12 driver. That driver is what Microsoft ships as the OpenGL Compatibility Pack (GLon12), so this is a real user configuration, not only a VM artifact: machines without a vendor OpenGL driver get it from the Store.

Reproduced on the win11 VM against the build from main (216ebd8), with `DUSKSTUDIO_OPEN_NOTEPAD=1` so no synthetic clicking is involved.

## What happens

Deterministic, 3 of 3 runs:

```
d3d12-plain#1 : DEAD exit=0x00000000   first frame completed
d3d12-plain#2 : DEAD exit=0x00000000   first frame completed
d3d12-plain#3 : DEAD exit=0x00000000   first frame completed
```

The stderr trail runs to the end of the first frame and stops:

```
GL vendor=Microsoft Corporation renderer=D3D12 (Microsoft Basic Render Driver)
   version=4.6 (Compatibility Profile) Mesa 26.2.0 glsl=4.60 maxTexture=16384
stage: open complete
stage: first onImGuiDisplay entered
... draw entered / style applied / window begun / ribbon / chart / status / window ended
stage: first onImGuiDisplay returned
<process gone>
```

`open()` completes, our whole draw completes, and the process dies in what follows: `ImGui::Render()` and `ImGui_ImplOpenGL3_RenderDrawData()` inside DPF's `onDisplay`.

## What is known

- Same binary, same VM, `GALLIUM_DRIVER=llvmpipe` (conforming GL 4.6, no D3D12): the notepad renders and keeps running. The notepad code is not at fault.
- A pre-GL3 display (GDI Generic 1.1) refuses cleanly and the DAW survives, so the capability check from #331 works.
- procdump on an earlier build caught `wil::ResultException` and `_com_error` (both `E06D7363`, MSVC C++ exceptions) with the process exiting `0x80070057`, which is `E_INVALIDARG` from a COM call inside the D3D12 layer.
- Since #331 the exit code is `0x00000000`. A clean exit means a graceful JUCE shutdown, so the process is being asked to quit rather than killed. For a DocumentWindow application that points at the main window being closed.
- The fatal frame is painted before the notepad's own timer ticks (no "first idle" line), so it arrives from the host message loop through pugl's WndProc. `pumpEvents()` never sees it, which is why its guard does not catch it.
- pugl does not call `PostQuitMessage`, so the obvious child-window-quits-the-host bug is not the cause.

## Next step

Branch `diag/windows-notepad-quit` adds two lines, one in `MainWindow::closeButtonPressed` and one in `DuskStudioApp::shutdown()`. Build 32544502472 is the artefact. If the close path runs, the child's failure is destroying the host window and the fix belongs in the embedding. If only `shutdown()` runs, something calls quit directly.

## Options if the call cannot be avoided

1. Find the GL call GLon12 rejects and stop issuing it, or work around it in the Dusk-owned DPF fork.
2. Refuse the notepad on that renderer the way a pre-GL3 display is refused. Blunt, and it costs GLon12 users the notepad, but it costs them the DAW today.

Doing nothing is not safe: the user loses unsaved work in the whole application, not just the notepad.

### LINKED ISSUE #333: [0.13.1 blocker] Strip the temporary notepad stage markers before tagging
#331 merged with the temporary stage instrumentation still in it. There are 15 `notepadLog ("stage: ...")` calls in `src/ui/NativeNotepadWindow.cpp`, tracing `open()`, the first idle, and every band of the first frame. They exist to place the Windows failures and should not ship.

They are stderr only and fire once per notepad open, so nothing is broken by them, but a release build should not narrate its own draw.

## Remove

- The `stage:` markers in `open()` (window construction, embedded offset, context, widget, font atlas, focus, open complete).
- The `stage: first idle` / `first idle returned` pair in `timerCallback`, and the `loggedFirstIdle` / `loggedFirstIdleReturn` members that drive them.
- The `stage: first onImGuiDisplay entered` / `returned` pair in `EditorWidget::onImGuiDisplay`, and `loggedFirstDraw`.
- The `mark(...)` lambda and its calls through `draw()`, and `tracedFirstDraw`.
- The corresponding resets in `open()`.

## Keep

- `setvbuf (stderr, nullptr, _IONBF, 0)` in `src/Main.cpp`. Windows leaves a redirected stderr fully buffered, and without this a process that dies without unwinding discards its log. This is what made the failures diagnosable at all.
- The `notepadLog` helper, including the `fflush` and the `OutputDebugStringA` mirror.
- `logGraphicsIdentity()` and its one line of GL vendor, renderer, version, GLSL version and max texture size. That is the first thing any Windows graphics report needs.
- The real diagnostics: "display provides no OpenGL 3 context", "cannot embed: ...", "graphics driver failed during the event pump".

Blocked by #332: the markers are what that investigation reads. Strip once it is settled.

Also retire branch `diag/windows-notepad-quit` when done.

### LINKED ISSUE #334: [0.13.1 blocker] Verify the 0.13.1 macOS DMG launches before tagging
0.13.1 exists because the 0.13.0 DMG could not launch on any machine (#327, fixed by #328). That fix has never been exercised against a packaged 0.13.1 DMG, so the release cannot be tagged on the strength of green CI: 0.13.0 passed every workflow and every asset check and still shipped an application that would not start.

The build and the test suite both run against the build tree on a machine where `/opt/homebrew` exists, so they pass whether or not the packaged bundle carries its dependencies. Only opening the artefact catches this.

Run on the M3 Air against a DMG built from the release commit, per `/release` Step 10:

```bash
MNT=$(mktemp -d)
set +o pipefail
yes | hdiutil attach -nobrowse -readonly -noverify -noautoopen -mountpoint "$MNT" "$DMG" >/dev/null
mount_rc=${PIPESTATUS[1]}
set -o pipefail
[ "$mount_rc" -eq 0 ] || { echo "STOP: cannot mount $DMG"; exit 1; }
APP="$MNT/DuskStudio.app"

codesign --verify --strict --deep --verbose=2 "$APP"

while IFS= read -r macho; do
  file -b "$macho" | grep -q 'Mach-O' || continue
  if otool -L "$macho" | tail -n +2 | awk '{print $1}' \
     | grep -vqE '^/usr/lib/|^/System/|^@executable_path/\.\./Frameworks/|^@loader_path/'; then
    echo "STOP: $macho links outside the bundle"; exit 1
  fi
done < <(find "$APP" -type f)

env -i HOME="$HOME" "$APP/Contents/MacOS/DuskStudio" --version
```

Checks that must pass:

- Every Mach-O in the bundle, not only the main executable, resolves inside the bundle. The relocated dylibs and the plugin-host helper are the ones that regressed.
- The signature verifies on the installed copy, after the install step rather than before it.
- The DMG contains `DuskStudio.app`, `LICENSE` and `LICENSES.txt` and nothing else. A `share/` tree means the Linux install rules leaked back in, which #326 fixed.
- The application actually opens, with a session, not just `--version`.

Note the reason `hdiutil` needs `yes |` and its own status read: the DMG carries a license agreement, and reading the pipeline status instead fails the step when `yes` dies of SIGPIPE under `pipefail`.

marc-korte:
Closes #332
Closes #333
Closes #334

Milestone 0.13.1 carried three blockers. All three are fixed here and verified against packaged artifacts rather than CI status, which is the check 0.13.0 lacked.

## #332 - Mesa D3D12 ends the application

Microsoft's OpenGL Compatibility Pack exposes Mesa's GL-on-D3D12 renderer. It advertises OpenGL 4.6 but terminates the host process while presenting the notepad's first ImGui frame, so the capability and version check from #331 is not enough on its own. `assessGraphicsCompatibility` now refuses that renderer by name. Native AMD, Intel and NVIDIA drivers do not identify their renderer as D3D12, and Mesa llvmpipe stays eligible.

Two exception paths are also guarded, because either one takes the whole application down rather than just the notepad:

- An exception escaping the DGL window constructor runs inside a native window procedure, where it is a fatal callback exception rather than something the caller can act on. `open()` converts it into the `false` return the caller already handles.
- An exception out of `app.idle()` reaches the host message loop the same way. The pump catches it, closes the notepad alone, and leaves the session holding the text the editor has already mirrored to it.

## #333 - temporary stage tracing

The 15 `stage:` markers, the `loggedFirstIdle` / `loggedFirstDraw` / `tracedFirstDraw` members and the `mark()` lambda are gone. Kept, per the issue: the unbuffered stderr in `Main.cpp`, the `notepadLog` helper, `logGraphicsIdentity()`, and the three actionable failure diagnostics. `tests/notepad_graphics_compatibility.cpp` asserts the tracing cannot come back. Branch `diag/windows-notepad-quit` has been deleted.

## #334 - the macOS disk image

Verified against a packaged 0.13.1 DMG on Apple Silicon: the signature verifies on the installed copy, all fifteen Mach-O files in the bundle resolve inside it, the image root holds only `DuskStudio.app`, `LICENSE` and `LICENSES.txt`, and the application opens a session and embeds the notepad from a scrubbed environment. CPack's `/Applications` symlink is disabled so the image root matches the packaging contract.

## Retina notepad sizing

The notepad takes its embedded geometry from the native view's backing scale on macOS. Cocoa peer coordinates are points and the framework's platform scale is 1 there, while DPF sizes the child in backing pixels, which left the editor at half size on a Retina display.

## No new JUCE

The scale seam and the rectangle math move into `NativeEditorEmbedScale.h` as framework-free functions, and the backing-scale query takes a native handle rather than a peer. That keeps every touched file at or below its recorded ceiling with no allowlist hand-edit: the tree goes from 9221 to 9206 occurrences. The allowlist update also captures slack left by earlier merged work in five unrelated files.

## Verification

- `tools/juce-gate.sh` passes; 9206 occurrences, ratcheting down.
- `cmake --build build -j5` clean, no new warnings.
- 734/734 ctest.
- macOS self-test 14 PASS / 0 FAIL, both from the build tree and from the installed copy off the DMG.
- Windows 11, MSI built from this branch, on the Mesa D3D12 renderer from the Compatibility Pack:

```
renderer=D3D12 (Microsoft Basic Render Driver) version=4.6 (Compatibility Profile) Mesa 26.3.0-devel glsl=4.60
Mesa D3D12 (OpenGL Compatibility Pack) is known to terminate the host; notepad unavailable
appeared=True  survived=True
```

  That configuration ended the process on 3 of 3 runs before this change. The same MSI refuses a GDI Generic 1.1 display cleanly and keeps the session open.

## Not covered

#335 stays open. The notepad still has no run on real Windows GPU hardware: the test VM can reach only GDI Generic 1.1 and Mesa D3D12, both of which this change refuses by design, so it cannot demonstrate a rendering notepad. The release notes already say the notepad is unverified on Windows GPU hardware.

A user on the Compatibility Pack sees the generic status message "Unable to embed session notepad with the current display backend"; the actionable reason reaches stderr only. Making it specific needs `open()` to return a reason rather than a bool, so it is left for a follow-up. `MANUAL.md` documents the refusal and the remedy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed notepad sizing on Retina Macs so embedded editors retain their proper dimensions.
  * Improved Windows protection against incompatible graphics drivers, including unsupported OpenGL compatibility layers.
  * Removed temporary startup and rendering diagnostics.
* **Packaging**
  * Simplified macOS disk images by removing the default Applications shortcut.
  * Updated packages and release information for version 0.13.1.
* **Documentation**
  * Added Windows graphics compatibility guidance and updated release notes.


marc-korte:
Review findings addressed.

**Float assertions in `tests/notepad_graphics_compatibility.cpp`** - valid, fixed in acb96df. The four `== Approx(...)` comparisons are now `CHECK_THAT` with `WithinRel`, and the file includes `catch2/matchers/catch_matchers_floating_point.hpp`. `catch_approx.hpp` and the `Catch::Approx` using-declaration are gone. This one contradicted the repo's own rule that float comparisons never use `==`. Tolerance is 1e-12: the values are exact products.

**`setEmbeddedGeometry` ignoring `geometry.scaleFactor`** - a real defect, not fixed here, filed as #338.

It is not implementable as described. DGL has no runtime scale setter: `Window` exposes `getScaleFactor()` and a virtual `onScaleFactorChanged(double)`, but the factor is fixed in the constructor. "Refresh the native child's DGL configuration" therefore means destroying and recreating the embedded window, which discards undo history, focus and in-progress chord entry.

The reachable half is the font atlas, and it is not safe to land unverified. `buildFontAtlas` calls `io.Fonts->Clear()` and `Build()`, and the ImGui OpenGL3 backend holds a font texture that has to be invalidated with it or the next frame draws from a stale atlas. That failure mode is blank or corrupted text, the same class as #331 and #332, and the scenario that triggers it (a Retina to non-Retina move, or a live UI-zoom change) is not something this branch has been able to exercise.

It is also pre-existing: `setEmbeddedGeometry` is byte-identical on the commit before this PR, and this PR does not touch it.

This branch exists to clear the 0.13.1 blockers and every change in it has been verified against a packaged artifact. Adding an unverifiable rendering change to it is the risk profile that produced 0.13.0, so it goes to #338 with the analysis and a suggested approach.

PR #339, opened 2026-08-23
Refuse the notepad after a driver ends the application drawing its first frame

marc-korte:
Targets 0.13.1.

Refusing a renderer by name, which is what #332 landed, only protects against drivers somebody has already lost a session to. The D3D12 failure exits zero through a graceful shutdown:

```
d3d12-plain#1 : DEAD exit=0x00000000   first frame completed
```

so there is no exception for the event-pump guard in #337 to catch, and nothing distinguishes an unknown driver failing the same way until it has already taken a session with it. #335 cannot close that gap by testing, either: three vendors and many driver versions, and a pass on one says nothing about the others.

## The guard

`FirstFrameProbe` writes a marker naming the renderer before the first frame is pumped, and removes it once one has completed. The covered window is exactly the single `app.idle()` call the failure happens inside. A run that never comes back leaves the marker, and the next launch declines to open the notepad:

```
[Dusk Studio/notepad] a previous run ended while the notepad was drawing its first frame on
Pretend Vendor GPU 9000; notepad unavailable. Delete <path> to try again
```

The refusal stands until the marker is removed, so a bad driver costs one session rather than every session. An unusable config directory costs the protection, not the notepad.

## The switch

`notepad_enabled` in `app-config.properties`, following the existing `getScanPluginsOnStartup` pattern.

The two are deliberately independent: the notepad owns the marker, the configuration owns the switch, and neither writes the other's state. Nothing cross-couples, so neither can wedge the other.

## Verification

macOS, against the real renderer:

- Healthy run: alive at 30 s, `GL vendor=Apple renderer=Apple M1 Pro version=4.1 Metal`, no marker left behind. That absence is the proof `disarm()` ran, because `arm()` is unconditional on the supported path and SIGTERM executes no cleanup.
- Planted marker: notepad refused by renderer name and path, application alive, marker persists so the refusal holds.

738/738 ctest, including four new cases: the clean arm/disarm cycle, a run that never came back staying refused across two launches, a marker with no renderer recorded, and no config directory at all.

`tools/juce-gate.sh` unchanged at 9206. The probe is framework-free and `NativeNotepadWindow.cpp` stays JUCE-free.

One honest limit: the arm-to-disarm window is about 16 ms, so an armed marker was not observed by racing the live application. That half rests on the unit tests and the straight-line code path.

## Not covered

The test VM cannot exercise arming: both renderers it can reach are refused before the marker is written, D3D12 by name and GDI Generic by version. The Windows check is the stale-marker refusal, which is what validates the `%APPDATA%` path and the message text.

#335 stays open. This does not verify the notepad on GPU hardware; it makes not having verified it survivable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved notepad recovery after graphics-driver failures during startup.
  - Failed renderers are identified, and the notepad remains disabled on the next launch to prevent repeated application failures.
  - Successful startup clears the recovery state automatically.
  - Failed openings now provide a more specific explanation when available.

- **New Features**
  - Added a setting to enable or disable the notepad.
  - Users can retry after changing graphics drivers by removing the recovery marker.

- **Documentation**
  - Updated release notes and system requirements with the new recovery behavior.


marc-korte:
Both findings verified against the code and fixed in cf3774e.

**Release notes overclaimed the guard.** Valid. The marker is cleared once the first frame completes, so a driver that fails on a later frame is not covered, and "while the notepad is drawing" read as though it were. Now scoped to "while the notepad is drawing its first frame".

**The refusal reason never reached the user.** Valid, and it mattered most for the case this PR adds. `open()` logged the renderer and the marker path to stderr and returned `false`; the status bar then said the display backend was at fault whatever had actually happened. For the first-frame marker the recovery is deleting a named file, which is unreachable advice when it only exists on a stream a GUI user never reads.

`open()` now keeps the reason it refused, phrased for a reader, and `MainComponent::toggleNotepad` shows it. Every refusal path carries one, not just the marker:

- the marker: names the renderer that failed and the file to delete
- the Compatibility Pack: says the renderer ends the application and to install the vendor driver, which closes the gap I flagged on #332
- no OpenGL 3 context
- an embed failure, with the underlying error

`false` on failure is unchanged, and the previous generic message stays as the fallback when no reason was recorded. Implemented as `lastOpenFailure()` rather than a callback or a categorised enum: the caller only displays the string, so a category it would immediately flatten into text would be indirection with one consumer.

Verified: 738/738, `juce-gate.sh` unchanged at 9206 (`setStatusText` takes `const char*`, so surfacing this adds no framework reference), and the stale-marker run on macOS still refuses with the DAW alive. The status bar itself renders this path already - the Windows screenshot on #332 shows the old generic string in exactly that slot, so only the text differs.

Still outstanding on this PR: the Windows check of the `%APPDATA%` marker path and message. The test box dropped off the network mid-transfer; the MSI is staged and verified locally, waiting on the box.

marc-korte:
Valid, fixed in 08847ed.

`open()` rejects an unusable parent or geometry before it reached the point the reason was cleared:

```cpp
stopTimer();
destroyEmbeddedWindow();
if (nativeParent == 0 || geometry.width < 2 || geometry.height < 2)
    return false;          // returned before lastFailure.clear()
```

so that path would have shown whatever a previous attempt recorded. The clear now happens on entry, before anything can return, which keeps the reason tied to the attempt it describes. Nothing else moved: the previous-failure branch and `lastOpenFailure()` are unchanged.

Worth saying that this path is close to unreachable today, since `toggleNotepad` screens a null peer first and `notepadGeometryFor` clamps the extent to `max(2, ...)`. It is fixed anyway because the invariant is cheap to hold and a future caller has no reason to know about either guard.

743/743, `juce-gate.sh` unchanged at 9206, and the stale-marker run on macOS still refuses by name with the marker preserved and the application alive.

marc-korte:
Valid, and it defeated the guard outright. Fixed in 0fcee3f.

`~Impl()` cleared the marker whenever the event pump had not failed. `graphicsFailed` is only set by a failing pump, and it is assigned after the `previousFailure()` early return, so the run that had just been *refused because of a marker* fell into the clearing branch and deleted it.

The full sequence: open the notepad on a driver that ends the application, lose the session, relaunch, get correctly refused, quit normally. The marker is gone, and the launch after that opens the notepad on the same driver and loses another session. The guard protected against exactly one occurrence and then reset itself.

Reproduced before fixing, on macOS with a planted marker and a real Apple Event quit:

```
marker before             : present
(app refuses by name, then quits cleanly)
marker after clean quit   : DELETED
```

Worth recording why this got through: every earlier run of this scenario ended in SIGTERM or `Stop-Process`, neither of which runs destructors, so `~Impl()` never executed and the marker survived for the wrong reason. The Windows cases on #332 have the same shape and were passing for the same wrong reason.

Ownership is now explicit. `armedMarker` is set only where `open()` actually arms the probe, cleared on entry to every attempt, and teardown clears the marker only when this run wrote it:

```cpp
if (armedMarker && ! graphicsFailed)
    probe.disarm();
```

A marker from an earlier run is the refusal the next launch has to see; one left by a failed pump has earned its keep; one this run armed and then tore down before the first pump is the false positive the earlier fix was for.

Verified with clean quits, not signals:

```
pre-existing marker + clean quit   refused by name, marker PRESERVED
healthy run + clean quit           GL identity logged, no marker left behind
```

743/743, `juce-gate.sh` unchanged at 9206.

marc-korte:
Valid, fixed in the latest commit, though narrower than the previous one.

`disarm()` reports nothing when it fails:

```cpp
void disarm() const
{
    std::error_code ec;
    std::filesystem::remove (marker, ec);
}
```

That is deliberate, since losing the guard is better than failing the notepad over it. But it means a marker can outlive its removal at first-frame time. With the old condition, a pump failure later in that same run then set `graphicsFailed` and teardown declined to clear it, so the notepad stayed refused after a run that had drawn frames perfectly well. A double fault rather than a live bug, but the failure mode is the same false positive class as the teardown case.

The condition now states what the marker means rather than approximating it:

```cpp
if (armedMarker && (firstFrameConfirmed || ! graphicsFailed))
    probe.disarm();
```

A completed frame settles the marker whatever happens afterwards, which also mops up a silent disarm failure. Only a run that armed one and never drew keeps it, which is the case the next launch has to see.

Re-verified all three lifetimes, using a real quit rather than a signal where the destructor matters:

```
pre-existing marker + clean quit   refused by name, marker PRESERVED
healthy run + clean quit           GL identity logged, nothing left behind
healthy run + SIGTERM              nothing left behind (the pump had already disarmed)
```

743/743, `juce-gate.sh` unchanged at 9206.

PR #356, opened 2026-08-26
gui: move the audio settings and mastering panels to native views (G2 remainder)

marc-korte:
Finishes G2. The last three JUCE panels become native framework views:
AudioSettingsPanel (with DuskAudioDeviceSelector) and the two mastering
editors (MasteringEqEditor, MasteringLimiterEditor). All four classes are
deleted. Gate moves 172 files / 8,684 uses to 164 / 8,253.

## What changed

- src/ui/imgui/AudioSettingsView + AudioDeviceSelector: the settings panel
  drawn with the widget kit against the same session atomics, app config and
  device manager. The device selector ships with the panel rather than ahead
  of it: it is the audio block's five rows, not a panel of its own.
- src/ui/imgui/MasteringEqView + MasteringLimiterView: the mastering stage's
  EQ and limiter as inline framework children on the rectangles the stage
  lays out. New DuskPanelView::wantsPlate() (default true) lets an inline
  view decline the modal plate and fill its whole child.
- PanelControls: shared form styling (combo, tick box, slider) on JUCE's
  dark scheme colours, written out with provenance. The capture confirms the
  combo fill lands on the same pixels as the JUCE panel.
- UI-scale preview: the panel is pinned to the physical size it opened at
  while the DAW behind it rescales (pinnedChildGeometry in
  NativeEditorEmbedScale.h). setOwnedBodyScaleInvariant and
  refitOwnedBodyToHost are deleted with their only caller.
- Sub-modal handoff: MIDI Bindings, Self-Test and device-error alerts are
  JUCE modals that would land behind an opaque native child, so the panel
  closes, the modal runs, and dismissing it reopens the panel. MainComponent
  owns both modals now. Scroll position resets across the handoff; MANUAL
  updated.
- Screenshot harness: a native panel reads its own steady frame back as PPM
  and ScreenshotCapture pastes it into the JUCE snapshot. qg-02, np-08 and
  mm-02 regenerated.

## Deliberate deviation from the spec

The plan sketched "EmbeddedModal::beforeModalShown as a list" as the unlock
for the mastering editors. Not used: that hook fires for every
EmbeddedModal::show, including DuskComboBox popups, and the stage owns two
combos, so a close-my-children hook would cycle two GL contexts on every
dropdown, and the hook has no matching signal to bring anything back.
Instead each panel keeps a JUCE NativePanelProxy carrying kPluginEditorTag:
its bounds are where the child goes and its visibility is what
PluginEditorHider already toggles for native plugin editors, which restores
as well as hides and gives show/hide with the stage for free. The spec doc
is updated to record this, since G3 and G4 hit the same problem.

## Allowlist hand-raises

Five files gain JUCE, all bridge code on the JUCE side of the seam:

- MainComponent.cpp 456 -> 457: applyGlobalUiScale (the X11 peer round trip
  for an app-wide scale change) moves here from the deleted
  AudioSettingsPanel.cpp. Net for that code across the repo is -99.
- NativeEditorEmbedScale.h 12 -> 14: PinnedChild carries a JUCE rectangle
  and takes a JUCE component reference; it is the seam that keeps the
  settings panel at its opening size while the DAW rescales.
- MasteringView.cpp 148 -> 154, MasteringView.h 28 -> 32: the
  NativePanelProxy stand-in, the proxy-to-child geometry helper, and the
  harness entry point.
- ScreenshotCapture.cpp 26 -> 40: the PPM paste (JUCE Image/BitmapData/
  Graphics) and the step runner. Harness-only; dies with G5.

Eight files leave the allowlist (the four deleted classes), and
EmbeddedModal.h ratchets 68 -> 64 without a hand edit.

## Verification

- Full build clean, zero new warnings (the two that appear in touched files
  predate the branch and were verified present at 0e723a8).
- ctest 810/810 (795 at the G2 merge). New headless suites:
  imgui_panel_controls (9 cases, controls driven through ImGui's real input
  queue, including a tooltip-wrap regression test proven to fail on the old
  code) and imgui_mastering_views (6 cases: draw without writing to
  read-only params, header engages the section, no plate taken).
- tools/juce-gate.sh passes at 164 / 8,253; every added JUCE line on the
  branch sits in the five files above.
- Side-by-side captures for all three panels against their JUCE renders
  (settings grid row for row, combo fill pixel-identical; EQ curve, grid,
  labels, band dots, FFT pill and 13 knobs; limiter wells, drag lines,
  value boxes, atten ticks, LUFS box). Full capture harness run exits 0
  with no stray intermediates.
- Independent review round fixed on-branch: tooltip wrapping was a no-op
  (tooltips grew to full sentence width), a mastering panel stayed blank if
  its proxy came back during the close-in-flight window, stale comments
  naming the deleted classes, and two doc counts.
- CodeRabbit round fixed on-branch: the sub-modal Done/Close buttons now
  return to the settings panel like Esc does, and populateMainOutput no
  longer clobbers a stored master routing while no device is open (real
  data-loss path via the ALSA periods re-apply cycle).

## Known edges, deliberate

- Any embedded modal or combo popup over the mastering stage closes and
  reopens both native children (an embedded DGL window cannot hide), and
  the EQ's FFT toggle resets to on with the fresh view. Persisting the
  toggle means a session-schema change; wants a product call rather than a
  review commit.
- Leaving the mastering stage, the native children stay mapped for about
  three frames over the console. Same framework limit.
- A combo popup opened from inside the multiband comp plugin editor can be
  buried by the EQ or limiter child (that path passes
  hidePluginEditors=false). Structurally the same shape as two native
  plugin editors in the mixer; a real fix means teaching PluginEditorHider
  a subtree exemption, deferred.
- The limiter's Mode dropdown uses the shared form styling rather than the
  JUCE editor's own darker fill, for family consistency.

## Bench debts

- Live-Wayland pass on all six G2 panels (the three here plus comp editor,
  startup dialog, virtual keyboard) and comp knob feel. Everything above
  ran under Xvfb or headless mutter; no real seat delivered input.
- The mastering-stage GL-cycle edge above is the thing most likely to be
  noticed live.

## Summary by CodeRabbit

* **New Features**
  * Added native audio settings, mastering EQ, and mastering limiter panels.
  * Added interactive EQ controls, spectrum visualization, limiter metering, and audio-device configuration.
  * Improved panel scaling, modal handoff, visibility, and screenshot capture behavior.

* **Bug Fixes**
  * Corrected native-panel availability documentation and settings-panel restoration behavior.

* **Documentation**
  * Updated build guides, user manuals, screenshot guidance, project status, and maintainer documentation.

* **Tests**
  * Added headless coverage for mastering views and shared panel controls.


marc-korte:
@coderabbitai review

Issue #357, opened 2026-08-26
LV2 file-backed state: cur/prev rotation may hand restored plugins stale absolute paths


marc-korte:
Spun out of the #355 fix (native plugin state restore going live for the first time). Not a 0.13.2 blocker — needs a real file-backed LV2 sampler to verify and touch safely, so it is deliberately deferred.

## Concern

`Lv2Instance::saveState` rotates the per-track state directory (`remove_all(prev)`, `rename(cur -> prev)`, `create_directories(cur)`) **before** `lilv_state_new_from_instance` asks the plugin for its file paths (see `src/engine/lv2/Lv2Instance.cpp` around the rotate at ~:819-837 vs `loadState` + `absolutePathCb`/`statepaths::toAbsolute` resolving against `<stateDir>/cur` at ~:857-886).

At restore the plugin is handed absolute paths under `<session>/state/lv2/trackNN/cur/`. If the plugin stores those and the next save rotates `cur -> prev` before serialising, the paths the plugin holds are momentarily dangling (bytes now in `prev/`), and two saves later `prev` is removed. Whether lilv's copy step recovers depends on whether the plugin re-reports the stale absolute path or re-resolves it.

Before #355 this was only reachable via `reactivate()` (device/SR change); now it is the normal path on every session open, so a file-backed LV2 sampler is the case to test.

## Suggested approach

Serialize into a fresh generation directory and rename it to `cur` only after `lilv_state_to_string` succeeds, instead of rotating first.

## Repro to build

A file-backed LV2 instrument (e.g. a sampler that references an external sample via LV2 state files), saved and reopened across two save cycles, checking the sample still resolves.

Issue #347, opened 2026-08-25
Plugin VST Sync problem  by  search as deafault


Brainzdj:
Hi,  
On Patreon, which I support to help you continue, I’ve seen the new version, but I can’t access it due to a VST error on my side. It can run the error DLL files, but it gets stuck during installation because there’s no option to skip. This mostly happens with plugins like Ujam, so please review it and let me know. Thanks.

marc-korte:
Hi @Brainzdj 

I am taking a look at this now and I will send an update soon. Thanks!

Brainzdj:
Thank you !

marc-korte:
Please try the latest version and let me know how it goes!

PR #358, opened 2026-08-26
Fix/0.13.2

marc-korte:
## Summary by CodeRabbit

- **New Features**
  - Plugin scans can now be cancelled, including with Escape, with clear completion status.
  - Startup scans show progress and results without interrupting alerts.
  - Native settings, startup, and virtual-keyboard panels provide improved controls and focus handling.

- **Bug Fixes**
  - Improved restoration of saved plugin session state.
  - Added safeguards for stalled or unresponsive plugins, including timeouts and quarantine handling.
  - Restored accurate latency compensation after loading plugin state.

- **Documentation**
  - Updated guidance for scanning, cancellation, MIDI controls, virtual keyboards, and compressor gestures.

PR #359, opened 2026-08-26
0.13.2: restore native plugin state on session reopen (#355) and unwedge the plugin scan (#347)

### LINKED ISSUE #355: Plugin's states are not saved or restored after saving and closing the session
OS + version: Manjaro Linux
Dusk build: 0.13.1
Audio backend: Pipewire
What you did: using my plugins, loading their presets etc... but their states are not restored after saving and closing the session in DuskStudio
Expected vs. what actually happened: 

I expected that the plugin's states are restored while re-opening a saved session.

To reproduced: 
  1- Set a MIDI track and insert Sunset Circuits plugin (I am sure you have this plugin haha :D )
  2- Change the synth model to "Modular" (for example)
  3- Save the session
  4- Close DuskStudio
  5 - Re-open the saved session and re-open Sunset Circuits plugin window: Sunset Circuits is back to its default state with the "Cosmos" synth.

This happened with my other plugins (not only Sunset Circuits). It happens with LV2, VST3 or CLAP plugin's formats as well.

marc-korte:
Fixes #355 and #347 for the 0.13.2 point release, cut from the v0.13.1 tag. Targets release/0.13 because main has moved on to de-JUCE work that must not ship in 0.13.x.

## #355 - plugin state never restored

Native-host plugin state was written to session.json with juce::Base64 (RFC 4648) and read back with juce::MemoryBlock::fromBase64Encoding, a size-prefixed private format that rejects every RFC string. The decode failed 100% of the time, so loadState was never called for any CLAP, LV2, VST3, AU or multisample slot, on tracks and aux lanes alike. The state on disk was valid all along.

- New src/foundation/Base64.h provides a JUCE-free RFC 4648 decoder; the two readers now use it. Encoders untouched. The JUCE-hosted PluginSlot path uses the MemoryBlock pair on both sides and is untouched.
- The 0.13.1 AU/multisample migrations copied MemoryBlock-encoded strings onto native keys verbatim. Those now transcode at migration time (src/engine/LegacyStateBase64.h), and the reader falls back to the MemoryBlock form so sessions 0.13.1 already migrated still restore. The fallback is unambiguous: that form always contains a '.', which RFC output never does.
- ClapInstance re-reads CLAP_EXT_LATENCY after loadState so PDC tracks a state-dependent latency (VST3 signals kLatencyChanged, LV2 has a latency port, AU is polled; CLAP was the gap).
- An unreadable blob now logs one stderr line instead of silently restoring defaults.

Verified end to end on Linux: mutated a parameter inside a real session's CLAP state blob, reopened, saved - the mutated value came back from the live plugin instance, which is only possible through decode -> loadState -> saveState. LV2 and VST3 slots load and round-trip; missing plugins degrade to the existing missing-plugin report.

## #347 - plugin scan wedges with no way out

The out-of-process scanner already had a 30s timeout, a blacklist and a dead-man's pedal, but ChildProcess::readProcessOutput blocks until the buffer fills or the child exits, so the deadline could never fire against a plugin sitting on a license dialog (Ujam and friends). The scan modal also had no cancel of any kind.

- A watchdog thread now owns every kill (abort or 30s deadline); killing the child is what releases the blocked read. Reap and kill are interlocked under one mutex so a kill can never land on a reaped (reusable) pid. A complete payload outranks the deadline, so a plugin that finished at 29.9s cannot be quarantined by the race.
- The same fix is shared with the native CLAP/VST3 bundle scan, which had an identical blocking-read loop and no abort at all (reproduced: a hanging .vst3 wedged dusk-studio-plugin-host --scan-native indefinitely).
- The scan modal gets a Cancel button and Esc support at both call sites. Cancel skips the in-flight plugin without blacklisting it and stops the remaining phases promptly; a cancelled manual scan says so in the completion alert instead of claiming success.
- Scan output is capped at 8 MB so a chattering child cannot grow the parent unboundedly.

Verified on Linux with a fabricated plugin that blocks in its loader: the JUCE phase quarantines it at 30s, the native VST3 phase (which previously hung forever) times out and skips it at 30s, the scan completes, and Esc mid-scan unwinds within a second or two.

Verified on the Windows 11 VM with the ci-windows-dev MSI built from this branch (7d128f8): silent install clean; a hand-built VST3 DLL that blocks in its entry point is quarantined by the JUCE phase at 30s and timeout-skipped by the native VST3 phase at 30s (TerminateProcess releases the blocked read as designed); the scan completes and the app stays healthy; clicking Cancel during the native-phase hang closes the modal within seconds and the startup log reports the scan as cancelled.

## juce-gate allowlist

--update recorded three decreases (AudioEngine.cpp 140 -> 139, RegionEditActions.cpp 36 -> 35, PluginManager.cpp 112 -> 109). Three hand edits need justifying:

- src/engine/LegacyStateBase64.h (new, 8): must call JUCE's own MemoryBlock base64 pair - reimplementing its bit-packing inside a data-preservation path is not a risk worth taking in a point release, and isolating it in a header is what made it unit-testable.
- src/ui/PluginScanModal.cpp 28 -> 33, .h 10 -> 11: the Cancel button is a juce::TextButton styled exactly as BounceDialog and FreezeDialog style theirs. There is no Dusk button class and the GUI tower has no seam yet.

## Known limitations (deliberate)

- A scan child whose grandchild inherits the pipe write end and survives the kill can still block the reader; no clean fix without fd-level access.
- LV2 and AU native discovery read manifests/registry only, cannot hang, and take no abort; cancel gates between phases.
- A cancelled native CLAP/VST3 pass is abandoned, not committed half-finished (committing a partial run would delete the missing rows from the picker and cache).
- LV2 file-backed state cur/prev rotation vs restored absolute paths is deferred to #357.

## Tests

764 ctest cases pass (9 new: RFC decoder round-trip/vectors/rejections, legacy transcode incl. the "0." empty-state form, 0.13.1-migrated fallback, session round-trip now decoding what it stores). Xvfb integration selftest 15/15. The migration helpers themselves are file-static in AudioEngine.cpp and stay untested directly; the transcode/fallback logic they call is what the new unit tests pin.


marc-korte:
Superseded by #360. This PR's head (fix/0.13.2) had main merged into it by a parallel session, so it carried all of main's de-JUCE / GUI-tower work (118 files) into the release path. #360 is cut clean from the v0.13.1 tag with only the two fixes + the review follow-up, and does not contain main. The fixes still land on main separately via #358.

PR #401, opened 2026-08-29
fix: preserve LV2 generations without resyncing shared state

marc-korte:
## Summary by CodeRabbit

* **Improvements**
  * Improved LV2 state saving by synchronizing only files that have changed, reducing unnecessary processing when many existing state files are present.
  * Enhanced state-generation handling for more reliable save and commit operations.

* **Tests**
  * Added coverage for shared state synchronization, including scenarios with large numbers of pre-existing files.

* **Documentation**
  * Updated the documented Catch2 test count from 874 to 875.


marc-korte:
@coderabbitai review

PR #474, opened 2026-09-03
fix(hosting): deliver transport panic reliably

### LINKED ISSUE #460: [Transport] Stop panic still misses muted MIDI tracks, unmapped VST3 synths and multi-port CLAP instruments
Found in the 0.13.3 pre-tag audit. #402 is fixed for the reported CLAP case; three routes still hold voices after Stop.

1. **Muted or un-soloed MIDI track.** `src/dsp/ChannelStrip.cpp:965` returns before the instrument runs when `!passByGate`, so the panic built at `AudioEngine.cpp:5045-5057` never reaches the synth. Mute a MIDI track mid-play, Stop, un-mute: the held voices resume. The silent-skip at `:1005` already exempts `isMidi`; the gate skip needs the same exemption (run the instrument, discard the audio).
2. **Native VST3 with no IMidiMapping for CC 120/123.** `Vst3Instance.cpp:672` delivers CCs only through `IMidiMapping`; `queueCcParam` returns at `:637` when unmapped and the panic is dropped. A VST3 synth that maps only mod and sustain keeps sounding. Synthesise `kNoteOffEvent` for the tracked sounding notes when the CC is unmapped.
3. **CLAP note ports above 0.** `ClapInstance.cpp:485` hardcodes `port_index 0` for `CLAP_EVENT_NOTE_CHOKE`. Instruments with several note-input ports keep voices on the other ports. Emit one choke per declared note-input port.

marc-korte:
Keep MIDI instruments processing behind mute and solo gates, synthesize VST3 note-offs when panic controllers are unmapped, and fan CLAP chokes out across every declared note port.

Closes #460

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ canonical source.
 
 ### Fixed
 
+- **Stop now silences every hosted instrument reliably.** Transport panic events
+  still reach MIDI tracks while they are muted or excluded by solo, VST3 synths
+  that do not map the standard panic controllers receive note-offs for their
+  sounding notes, and CLAP instruments receive a choke on every note-input port.
 - **The macOS release now includes the plugin sandbox.** The disk image kept
   its macOS 11 deployment target, but the sandbox's shared-address wake API
   required macOS 14.4, so CMake omitted the plugin-host helper and scans ran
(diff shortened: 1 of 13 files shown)


marc-korte:

PR #475, opened 2026-09-03
fix(session): clone native inserts on Linux

### LINKED ISSUE #462: [Linux] Clone Track drops native CLAP/LV2/VST3 inserts: the native capture and replay are compiled only under DUSKSTUDIO_HAS_NATIVE_AU
Found in the 0.13.3 pre-tag audit. `src/session/RegionEditActions.cpp`.

The capture at `:635` and the replay at `:847` are both wrapped in `#if DUSKSTUDIO_HAS_NATIVE_AU`, with the CLAP / LV2 / VST3 blocks nested inside (`:638-678`, `:856-914`). The comment at `:847` says "This is macOS-only (the AU gate), so Linux clone behaviour remains byte-for-byte unchanged". The effect on Linux, the primary platform:

- cloning a track with a native CLAP, LV2 or VST3 insert produces a clone without it, silently
- the destination track's existing native insert is not cleared, so a clone can carry a stale plugin
- the `retainStateForLiveIdentity` calls added for #389 are dead code on Linux
- `:982` commits the primed multisample even when `primed.stateRestoreFailed`, installing a live wrong-preset instrument while the slot is flagged offline, contrary to `NativeMultisampleSlot.h:96-102` and the engine's own restore gate

Guard each format on its own `DUSKSTUDIO_HAS_NATIVE_*` macro, gate the multisample commit on `msStateRestored`, and add a Linux clone round-trip test for a native insert (the CLAP fixture is already built by CI).

marc-korte:
Closes #462

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ GPL source on this repo — build from source and the binary costs you nothing b
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 905 Catch2 test cases across 177 test source files. Linux
+The C++ suite declares 906 Catch2 test cases across 177 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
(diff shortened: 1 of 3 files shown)


marc-korte:

PR #476, opened 2026-09-03
fix(pdc): ignore quarantined insert latency

### LINKED ISSUE #461: [PDC] recomputePdc credits the latency of a quarantined native insert that is passing dry audio at zero delay
Found in the 0.13.3 pre-tag audit. Introduced by the reactivation policy work (PR #398, `NativeRestorePolicy.h`).

`src/engine/AudioEngine.cpp:1027-1045` reads each native insert's latency gated only on `!isBypassed()`. `NativeInsertSlot::quarantineAfterFailedReactivation` (`NativeInsertSlot.h:119`) leaves the instance loaded, clears `processingOnline`, and the process path (`:195-205`) copies input to output at zero delay. `isBypassed()` is untouched, so PDC still delays every other track by the quarantined plugin's latency.

A 2048-sample linear-phase VST3 that fails reactivation on a sample-rate switch plays its track about 21 ms early against the mix, with no indication beyond the stderr line from the quarantine.

Gate each latency read on `slot.isProcessingOnline()` as well as bypass, and recompute PDC from the quarantine and the later successful reactivate.

marc-korte:
Closes #461

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ GPL source on this repo — build from source and the binary costs you nothing b
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 905 Catch2 test cases across 177 test source files. Linux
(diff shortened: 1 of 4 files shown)


marc-korte:

PR #478, opened 2026-09-03
fix(midi): preserve dense output blocks

### LINKED ISSUE #465: [MIDI out] Per-track MIDI-out scratch is a quarter of the input capacity, and copyDuskMidiSorted drops the whole block on overflow
Found in the 0.13.3 pre-tag audit. `src/engine/AudioEngine.cpp`.

`:3009` reserves `perTrackMidi` at `4 * kMidiBlockBytes`; `:3013` reserves `midiOutTrackScratch` at `1 * kMidiBlockBytes`. `copyDuskMidiSorted` (`:36` in the helper) clears the destination and gives up when it fills, so a dense MIDI region on a track routed to a hardware MIDI output loses entire blocks, note-offs included. External gear is left with stuck notes.

Also: `copyDuskMidiSorted` (`:46`) is an O(P*N) selection sort on the audio thread once per MIDI-out track per block. Near the 64 KB cap that is on the order of 10^7 inner iterations inside the callback. Reserve the scratch at the same size as the source and sort an index array with `std::sort` over fixed storage.

marc-korte:
Closes #465

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ GPL source on this repo — build from source and the binary costs you nothing b
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 904 Catch2 test cases across 177 test source files. Linux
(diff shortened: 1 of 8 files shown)


marc-korte:

PR #479, opened 2026-09-03
fix(audio): publish device callback snapshots

### LINKED ISSUE #464: [RT] DeviceManager fan-out takes a std::mutex on the audio thread and holds it across device start
Found in the 0.13.3 pre-tag audit. `src/engine/device/DeviceManager.cpp:87`.

`CallbackFanout::audioDeviceIOCallback` locks `callbackListLock` (`std::mutex`) on every block. CLAUDE.md's audio-thread rules forbid this outright. The same lock is held across `audioDeviceAboutToStart` -> `prepareForSelfTest`, which reloads every native plugin and writes to stderr, so any `addCallback` / `removeCallback` or device change blocks the RT thread behind a non-RT holder for as long as that takes. The symptom is an xrun on device change or on self-test, which is easy to misattribute to the backend.

Publish the callback list as an atomic snapshot (retire the old vector after the swap, the way `PluginSlot` retires instances) and keep the prepare path outside any lock the audio thread can touch.

marc-korte:
Closes #464

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ GPL source on this repo — build from source and the binary costs you nothing b
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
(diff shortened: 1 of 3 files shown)


marc-korte:

PR #480, opened 2026-09-03
fix(scan): honor LV2 and AU cancellation

### LINKED ISSUE #466: [Plugin scan] LV2 and AU phases ignore the Cancel/abort flag
Found in the 0.13.3 pre-tag audit. Follow-on to #347.

`src/engine/PluginManager.cpp:683` and `:685` check `aborting` and `cancelled()` before calling `scanLv2Plugins()` and `scanAuPlugins()`, but neither function takes the flag (`:887`, `:965`) and neither checks it per bundle. Once the scan reaches the LV2 phase, Cancel does nothing; `PluginScanModal.cpp:60` then force-stops the worker with `stopThread(5000)` mid-lilv, which is undefined behaviour at quit. Thread the abort flag through both and check it per bundle, the way the CLAP/VST3 phases do since #408.

marc-korte:
Closes #466

## Summary by CodeRabbit

* **Improvements**
  * Plugin discovery scans can now be cancelled while in progress for LV2 and Audio Unit plugins.
  * Cancelling a scan returns promptly without producing partial results.
  * Plugin discovery results remain stable when scans complete normally.

* **Tests**
  * Added regression coverage for cancelling LV2 and Audio Unit scans before processing begins.

### DIFF
--- a/src/engine/PluginManager.cpp
+++ b/src/engine/PluginManager.cpp
@@ -680,9 +680,9 @@ int PluginManager::scanInstalledPlugins (
     const auto cancelled = [abort] { return abort != nullptr && abort->load (std::memory_order_relaxed); };
 
     if (! aborting && ! cancelled()) scanClapPlugins (abort);        // CLAP isn't a juce format - scan it alongside the JUCE pass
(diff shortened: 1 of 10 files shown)


marc-korte:

PR #481, opened 2026-09-03
fix(windowing): honor foreground activation

### LINKED ISSUE #449: [Windows/macOS] bringWindowToFront grants foreground rights to itself; flushWindowOperations is still an empty stub
Found in the 0.13.3 pre-tag audit. Follow-on to #369 and #368.

**Windows.** `src/ui/PlatformWindowing_Windows.cpp:120` calls `AllowSetForegroundWindow(GetCurrentProcessId())`. That API grants foreground rights to *another* process; calling it with our own PID is a no-op. The primary instance never has the user's last input (the second instance does), so `SetForegroundWindow` fails and every handoff only flashes the taskbar. The unsaved-changes prompt stays buried. The fix is on the other side: the second instance calls `AllowSetForegroundWindow(primaryPid)` before writing the handoff payload, and the primary drops the self-PID call. `SingleInstance` needs to learn the primary's PID (the pipe server PID is available via `GetNamedPipeServerProcessId`).

**Both.** `flushWindowOperations()` is `{}` on Windows (`:133`) and macOS (`PlatformWindowing_Mac.mm:127`) while `PlatformWindowing.h:96-101` documents a deterministic message-pump / run-loop drain that shutdown relies on to space window destroys. Implement it or delete the contract from the header so the next reader is not misled.

The activation tests (`tests/platform_window_activation.cpp:91-160`) are source-text greps and cannot catch either.

marc-korte:
Closes #449

### DIFF
--- a/src/ui/PlatformWindowing_Mac.mm
+++ b/src/ui/PlatformWindowing_Mac.mm
@@ -124,7 +124,12 @@ void bringWindowToFront (juce::ComponentPeer& peer)
         [window deminiaturize:nil];
     [window makeKeyAndOrderFront:nil];
 }
-void flushWindowOperations()                                {}
+void flushWindowOperations()
+{
+    [[NSRunLoop currentRunLoop]
+        runMode:NSDefaultRunLoopMode
+        beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.001]];
+}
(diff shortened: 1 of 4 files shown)


marc-korte:

PR #540, opened 2026-09-09
fix(app): route termination signals through the staged shutdown

### LINKED ISSUE #507: SIGTERM bypasses the staged shutdown
On Linux the app has no SIGTERM handler. JUCE's message loop only installs a SIGINT handler, and `DuskStudioApp::systemRequestedQuit()` is a bare `quit()` that does not route through `MainComponent::requestQuit` -> `beginSafeShutdown`.

Result: a logout, `systemctl --user stop`, or any supervisor sending SIGTERM kills the process with exit 143 and none of the staged teardown runs. The unsaved-changes prompt never appears, plugin editors are not dropped before the plugin instances, and out-of-process children are left to the reaper.

Expected: SIGTERM posts `requestQuit()` onto the message thread (via `dusk::callAsync` from an async-signal-safe handler that only sets a flag or writes a pipe), so the same phase sequence runs as for a window close. `systemRequestedQuit()` should do the same rather than `quit()` directly.

Found while designing the scenario suite: a black-box "clean quit" leg cannot use SIGTERM today because it produces zero shutdown markers.

marc-korte:
Closes #507

## The bug

Nothing in `src/` set a disposition for SIGTERM, and JUCE's Linux loop installs a handler for SIGINT only (`juce_Messaging_linux.cpp:304`). `DuskStudioApp::systemRequestedQuit()` was a bare `quit()` that never reached `MainComponent::requestQuit`.

Reproduced against `origin/main` under a private Xvfb display: SIGTERM to the GUI process exits **143 with zero `[Dusk Studio/shutdown]` markers**. No unsaved-changes prompt, plugin editors dropped after the instances they belong to, sandbox children left to the reaper.

## The fix

`dusk::QuitSignalHandler` in the message-thread seam (`src/foundation/MessageThread.h/.cpp`):

- Traps SIGTERM, SIGINT and SIGHUP. The handler is async-signal-safe: it writes a `volatile sig_atomic_t` and nothing else. `dusk::callAsync` allocates and takes the message-manager lock, so it cannot be called from signal context.
- A 50 ms message-thread watcher consumes the latch and runs the callback.
- RAII. The destructor restores the default dispositions. `DuskStudioApp` owns one for the window's lifetime and resets it first thing in `shutdown()`.

`systemRequestedQuit()` now routes into `MainComponent::requestQuit` unless the staged shutdown is already running. That guard matters: `beginSafeShutdown` phase 7 posts a `systemRequestedQuit` of its own, and without the check the app would re-enter `requestQuit`, hit the `shutdownInProgress` latch, and never quit. `MainComponent::isShuttingDown()` exposes the latch for it.

Same phase sequence as the titlebar X, unsaved-changes prompt included.

## Verification

Linux (this box):

- SIGTERM under Xvfb before: `EXIT=143`, 0 shutdown markers. After: `EXIT=0`, all 8 phases printed.
- `cmake --build build -j3` clean, zero new warnings.
- `ctest --test-dir build-tests` 981/981 pass.
- `tools/juce-gate.sh` unchanged at 163 files / 8067 uses. The new code adds no JUCE.

macOS (M3 Air, arm64, clang):

- `cmake --build build -j5` succeeds, zero warnings from the touched files. The first Mac run surfaced two clang-only warnings in the new code (`-Wmissing-prototypes` on the signal handler, `-Winconsistent-missing-destructor-override`); both are fixed in this branch.
- `ctest --test-dir build-tests` 956/956 pass, new case included.
- No end-to-end SIGTERM run on the Mac: GUI launches from ssh abort in MainWindow on that box, and the handler is only installed on the GUI path, so the headless self-test would not exercise it. The unit test covers install plus latch there; the dispatch half runs on Linux and Windows.

## Test

`tests/foundation_message_thread.cpp` gains a case that installs the handler, raises a real SIGTERM, asserts the callback did **not** run from signal context, then pumps the loop and asserts it did. README's test-case count contract is bumped 1001 to 1002 to match.

## Manual

`MANUAL.md` "Save changes before quitting?" now says a logout, a machine shutdown, or stopping the app from a terminal counts as quitting and raises the same prompt.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -2286,7 +2286,7 @@ The format for each entry:
 
 ### Save changes before quitting?
 
-- **When**: You quit with unsaved changes.
+- **When**: You quit with unsaved changes. Logging out, shutting the machine down, or stopping the app from a terminal counts as quitting: the prompt appears then too, and the session waits on your answer.
 - **Text**: "Your session has unsaved changes since the last manual save. If you don't save, the autosave will still be available the next time you open this session."
 - **Buttons**: **Save** / **Don't Save** / **Cancel**.
 - **Action**: Save unless you specifically want to discard. The autosave file remains as a safety net regardless.
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1001 Catch2 test cases across 188 test source files. Linux
+The C++ suite declares 1002 Catch2 test cases across 188 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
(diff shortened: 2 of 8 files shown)


marc-korte:

PR #541, opened 2026-09-09
fix(oop): refuse PrepareToPlay with no plug-in and acknowledge the POSIX handoff

### LINKED ISSUE #503: [OOP] Small residues from the milestone-6 audits
Found in the 0.13.3 pre-tag re-audit and the milestone closeout. None blocks the tag; grouped so they are not lost.

- `src/util/SingleInstance.cpp:249`: the `<slot>.lock` companion file is never unlinked, one stray file per runtime dir.
- POSIX single-instance handoff has no acknowledgement: a second launch in the same millisecond as `kill -9` on the primary gets its connect accepted into the dying backlog, sees a completed write, quits, and the session path dies with it. Windows acks.
- `src/engine/PluginSlot.cpp:598-628` `releaseResources` sends Release to the child, and `prepareToPlay` (`:558-592`) then sends only PrepareToPlay, which `handlePrepareToPlay` answers with status 0 for a null instance. Only shutdown calls release today; wiring it into a device change would republish a child with no plugin.
- `src/ui/ChannelStripComponent.cpp:2748,2809`: `(void) hideRemoteEditor()` presents the shell even when the child refused to hide its fallback window, so both can be on screen.

marc-korte:
Closes #503. Covers residues 1, 2 and 3 of the four; residue 4 needs no change (verdict and reasoning on the issue: a failed hide means the child is wedged, crashed or gone, bailing never removes the stuck window and only withholds the working shell editor).

## Residue 3 - PrepareToPlay succeeded with no plug-in

`handlePrepareToPlay` returned status 0 when `host.ownedInstance` was null, so after a Release the parent's success path republished `currentRemote` and routed audio into a child with nothing to process. The slot reads as live and is silent, with no error anywhere.

Latent today because only shutdown calls `releaseResources`, exactly as the issue says; wiring release into a device change would surface it. PrepareToPlay with no instance now returns `kControlStatusNoPluginLoaded`, so the parent's existing failure path nulls `currentRemote` and leaves the slot bypassed.

Test connects to a real `--ipc-host` child and calls `prepareToPlay` without loading anything. No plug-in fixture needed, since the defect is precisely the empty-child case. Fails on main, passes here.

## Residue 2 - POSIX handoff had no acknowledgement

`handOver` treated a completed `writeAll` as delivery. A launch racing a `kill -9` on the primary gets its connect accepted out of the dying process's backlog and its write buffered, reports Delivered, and quits, taking the session path with it.

The listener now writes a one-byte acknowledgement after the payload is queued on its message thread, and the sender waits for it. An unacknowledged handoff returns the new `Handoff::Submitted`, which the existing caller already treats as "run unattached" rather than quitting. This is the shape Windows already had, including the `Submitted` state.

Test binds a listening socket that never accepts, so the kernel completes the connect and buffers the payload with no reader. Before: the launch quits. After: it runs unattached. Fails on main, passes here.

## Residue 1 - slot lock file - wontfix

The `<slot>.lock` companion is deliberately never unlinked, and this PR records why at the destructor rather than changing behaviour. Unlinking it while another launch holds it open leaves that launch holding `flock` on an unlinked inode while a third creates and locks a fresh inode at the same name, so both believe they own the slot. That is the double-primary the lock exists to prevent, and it is the same class of bug as residue 2. The cost of leaving it is one empty file per runtime directory, in `/run/user/<uid>`, which the session clears at logout.

## Residue 4 - no change needed

`ChannelStripComponent.cpp:2661` and `:2725` (the issue's `:2748`/`:2809` have drifted). The audit finding does not hold up. `hideRemoteEditor` returns false for four different reasons, only one of which is "a child window may still be up", and the child answers HideEditor successfully when it has no editor window (`destroyEditorOnMessageThread` no-ops on a null `editorWindow`). So a failure means the child is wedged, crashed or gone. A crashed child's window died with it; a wedged child's window stays up whether or not we present the shell. Bailing never removes the stuck window, it only withholds the working editor, so the existing comment is right on the merits. Full reasoning on the issue.

## Verification

- `ctest`: 982/982 pass. Both new tests verified failing on main and passing here.
- `tools/juce-gate.sh`: pass, 163 files / 8067 uses, unchanged.
- App builds clean at `-j3`, no new warnings.
- `scripts/run-selftest-xvfb.sh`: exit 0, 15/15. Never run on the live session.

README's declared test count moves to 1003 across 189 files because `release-mechanics-contract` asserts it. **This collides with PR #538, which moves it to 1002 across 188.** Whichever merges second needs a rebase and a recount; after #538 this becomes 1004 across 189.

No overlap with #538 otherwise: residue 3's fix is entirely child-side, so `PluginSlot.cpp` is untouched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin preparation now reports an error when no plug-in is loaded, instead of indicating success.
  * Single-instance command handoffs now confirm delivery; launches proceed independently when confirmation is unavailable.

* **Tests**
  * Added coverage for preparing without a loaded plug-in and unacknowledged single-instance handoffs.

* **Documentation**
  * Updated README test suite statistics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1002 Catch2 test cases across 188 test source files. Linux
+The C++ suite declares 1004 Catch2 test cases across 189 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
@@ -123,7 +123,7 @@ src/
   session/     # Session model + JSON serialisation
   ui/          # MainComponent, ConsoleView, channel/aux/master strips, mastering view
   util/        # Native log storage + CrashHandler signal reports
-tests/         # 1002 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
+tests/         # 1004 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
 packaging/     # .desktop, AppStream, MIME, macOS bundle — for tarball + DMG builds
 DuskStudio.md  # authoritative product spec
 MANUAL.md      # end-user manual (Pandoc-buildable to PDF via docs/build-pdf.sh)
--- a/src/engine/ipc/PluginHostMain.cpp
+++ b/src/engine/ipc/PluginHostMain.cpp
@@ -1231,7 +1231,7 @@ std::uint32_t handlePrepareToPlay (HostState& host,
     if (payload.size() < sizeof (PrepareToPlayPayload)) return 1;
     PrepareToPlayPayload p {};
     std::memcpy (&p, payload.data(), sizeof (p));
-    if (host.ownedInstance == nullptr) return 0;
+    if (host.ownedInstance == nullptr) return kControlStatusNoPluginLoaded;
     return withParkedHostWorker (host, [&]
     {
         host.ownedInstance->prepareToPlay (p.sampleRate, p.blockSize);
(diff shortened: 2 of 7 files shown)


marc-korte:

Issue #535, opened 2026-09-09
[SFZ] Local instrument browser for soundfonts already on disk


marc-korte:
## Context

Loading an instrument today means Open local file from the multisample editor and knowing where the file is. `src/ui/multisample/DuskMultisampleEditor.*` has the file browse path and the SF2 preset picker; there is no view of what the user already has.

The four-part downloadable SFZ plan (#74, #252, #253, #254) is deferred past 1.0. That plan is about fetching packs over the network. This issue is the offline half of it, which is what 1.0 needs: a browser for instruments already on disk.

Splitting it this way means no network code, no signed catalog, no archive extraction and no libcurl or libarchive dependency lands for 1.0.

## Work

- Add a library root setting, defaulting to a per-user instruments directory, with the ability to add more roots.
- Scan the roots for `.sfz` and `.sf2` on explicit user action, never at startup and never during session restore. Cache the result.
- Present an embedded modal listing what was found: name, format, file size, folder. Filter as you type.
- Load the selected instrument through the existing multisample load path. Keep Open local file.
- Report a root that has gone away, and an instrument file that has been deleted since the scan, as a clear state rather than a silent empty list.
- Reuse `DuskComboBox` and `duskstudio::showContextMenu`. No raw framework widgets.

## Out of scope

Anything that touches the network, the signed catalog, downloads, or archive extraction. Those stay on #74 and its children.

## Acceptance criteria

- The browser opens from the multisample editor and from the insert chooser.
- A directory containing a mix of `.sfz`, `.sf2` and unrelated files lists exactly the instruments.
- Typing filters the list, and clearing the filter restores it.
- Selecting an instrument loads it and the session saves and reloads that choice.
- Nothing scans on startup or during session load. Prove this with a test that fails if a scan is triggered from either path.
- Pointing a root at a directory that does not exist shows a stated error, not an empty list.

marc-korte:
Amending one acceptance criterion. The original said the browser opens from the multisample editor **and** the insert chooser. The insert chooser is the plugin picker, and putting a soundfont library entry into a list of plugins is a product decision rather than a mechanical one, so the picker route is deferred past 1.0.

For 1.0 the browser opens from the multisample editor only, beside the existing "Open local file" button, which stays as it is. Everything else on the list is unchanged.

[...]

PR #543, opened 2026-09-09
feat(sfz): scan local directories for soundfonts

marc-korte:
Refs #535. PR 1 of 3: the model and its tests. No UI, and nothing here touches the network, a catalog, an archive, libcurl or libarchive.

## What lands

`src/engine/sfz/SfzLibrary.{h,cpp}` is framework-free and built on `dusk::fs`. It walks a list of roots for `.sfz` and `.sf2` and returns entries (path, display name, containing folder, format, size) plus, for every root it could not use, a stated reason.

Reporting bad roots rather than dropping them is deliberate: a library that silently lists nothing looks exactly like one that is genuinely empty, and the user cannot fix what they cannot see.

The walk is depth-capped at 6 and carries a set of canonical directory identities. The cap alone does not stop a symlink loop, it just bounds how far each branch of it is walked, so identity is what actually terminates. The same set also means two roots that overlap, or one that symlinks into another, list their shared instruments once.

`scanLibraryRoots` blocks on the calling thread and takes a cancel flag polled per directory and per file, plus a progress callback fired once per root. The panel in PR 2 owns the worker. Nothing here belongs on the audio thread or the message thread.

`filterEntries` is a pure case-insensitive substring match over name and folder, so the panel can filter without rescanning and an empty query restores the full list.

## Default roots

One table per platform in `SfzLibrary.cpp`, all skipped without complaint when absent:

- Linux: `/usr/share/sounds/sf2`, `/usr/share/soundfonts`, `~/.local/share/sounds/sf2`, `~/.local/share/soundfonts`, `~/.local/share/sfz`, `~/soundfonts`
- macOS: `/Library/Audio/Sounds/Banks`, `~/Library/Audio/Sounds/Banks`, `~/Music/Soundfonts`
- Windows: `%USERPROFILE%\Documents\Soundfonts`, `%LOCALAPPDATA%\Soundfonts`

The first two Linux entries are where distribution packages install, so a fresh install carrying a packaged General MIDI bank shows something the first time the browser opens.

**This list is a first guess and is flagged for review.** It is one table in one file precisely so it can be corrected in one place.

## User roots

Two `appconfig` accessors. The store is one key per line, so a newline-separated list silently truncated at the first entry, which the round-trip test caught. Each root is now percent-encoded and the list joined with `:`, so a path containing a colon, a newline or a percent sign round-trips intact instead of splitting into roots that do not exist.

## Coverage

Eleven new cases: extension filtering, a missing root reported rather than swallowed, the depth cap, a symlink loop, overlapping roots, cancellation, progress ticks, case-insensitive filtering and clearing back, default roots absolute and distinct, and the two appconfig round-trips including the awkward-path one. The symlink case skips where the filesystem has no symlinks.

`ctest`: 991/991. `tools/juce-gate.sh`: 163 files / 8067 uses, unchanged; the model adds no framework tokens. App builds clean at `-j3`.

## Size

575 insertions, of which 204 are tests and 90 are the config accessors, leaving about 280 lines of model. Over the 400-line guide once tests are counted; splitting the tests from the code they cover would have been worse.

## Next

PR 2 adds the embedded-modal panel (and hand-adds its two files to the coupling allowlist, justified there). PR 3 covers MANUAL.md. The picker entry point is deferred per the design discussion; I am updating the acceptance criteria on #535 to say so.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -690,6 +690,7 @@ target_sources(DuskStudio PRIVATE
     src/engine/PluginDescriptor.cpp
     src/engine/PluginManager.cpp
     src/engine/PluginSlot.cpp
+    src/engine/sfz/SfzLibrary.cpp
     src/engine/RecordManager.cpp
     src/dsp/AuxLaneStrip.cpp
     src/dsp/BusStrip.cpp
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1001 Catch2 test cases across 188 test source files. Linux
+The C++ suite declares 1012 Catch2 test cases across 189 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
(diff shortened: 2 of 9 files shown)


marc-korte:

PR #544, opened 2026-09-09
feat(sfz): browse the soundfonts installed on this machine

marc-korte:
Refs #535. PR 2 of 3: the panel. Stacked on #543, so review that one first; this branch contains it.

MANUAL.md rides here rather than in a third PR, since the browser is user-visible the moment this lands.

## What the user gets

**Library...** joins **Browse...** in the soundfont editor header. It lists the `.sfz` and `.sf2` the configured roots already hold, filters by instrument or folder as you type, and hands the chosen file to the editor's existing `startAsyncLoad`. Nothing about loading or session storage changes: the choice still persists as `Track::nativeMultisamplePath`.

**Browse...** is untouched and still the way to reach a file the library does not cover.

Instrument library

Roots that are missing or unreadable head the list with their reason rather than being dropped, because an empty library and a mistyped root are otherwise indistinguishable. That path is visible in the figure history: the first capture attempt rendered exactly that row when the fixture directory did not exist.

## Threading

The scan runs on a worker the panel owns. `stopScan` raises the cancel flag and joins, and both the destructor and every fresh scan go through it, so a scan cannot outlive the panel or overlap another. It starts only when the panel opens, on **Rescan**, or after **Add folder...** — never at startup, never during a session load, never on the audio thread.

## Coupling allowlist

`src/ui/multisample/SfzLibraryPanel.{h,cpp}` are hand-added at 15 and 65. The GUI tower has no framework-free seam, so a new panel cannot avoid coupling; this is the case CLAUDE.md allows a hand edit for. Nothing else rose:

- `DuskMultisampleEditor.h` stays at 21. The new button is declared alongside `browseButton` in the same declaration, so a new member costs no new name.
- `DuskMultisampleEditor.cpp` **drops** 68 to 67. The click handler needed one `SafePointer`, offset by moving two `jmax`/`jmin` calls to `std::`, and the allowlist records the lower number.
- `ScreenshotCapture.cpp` stays at 40, with the fixture filenames built as `std::string` and the wait loop reusing the file's own `settle`.

Gate reports 165 files / 8146 uses, up from 163 / 8067: the two new files and nothing else.

## Screenshots

Captured under Xvfb through `docs/capture-screenshots.sh` and inspected. Three iterations were needed and each caught something real:

1. The panel sat on "Scanning..." forever. `settle()` sleeps rather than pumping, so the panel's own timer never fires during capture. Hence `applyFinishedScan()`, which the timer and the harness both drive.
2. The fixture rendered as a "not found" problem row, because `replaceWithText` does not create parent directories.
3. The status line was blank, because it only updated from the timer. Split into `updateStatus()`, called from both paths.

Only the new figure is committed. Re-running the harness rewrites every PNG with encoding noise; that churn is reverted so this PR carries one image.

## Verification

`ctest` 991/991. Gate as above. App builds clean at `-j3`. `docs/screenshot-list.md` gains its row.

## Not in scope

Network, catalog, downloads, archives. The plugin-picker entry point is deferred per the note on the issue; the library opens from the multisample editor only.


marc-korte:

PR #545, opened 2026-09-09
fix(windows): resolve the IPC harness child under its real name

### LINKED ISSUE #504: [Windows] IPC self-test harnesses resolve the child without .exe and hang
Found while running the 0.13.3 pre-tag build on the Windows 11 VM.

`DUSKSTUDIO_RUN_IPC_SELFTEST=1` and `DUSKSTUDIO_IPC_HOST_TEST` resolve the child as `exe.getSiblingFile ("dusk-studio-plugin-host")` (`src/DuskStudioApp.cpp:2456` and `:2472`), while real out-of-process loads use the `.exe` name from `src/engine/PluginManager.h:173`. On Windows the harness path does not exist, the connect never gets a child, and the process sits until killed: no output after the startup lines, both with the 0.13.3 candidate and with the 08/26 build installed on the VM, so this is not a regression of the milestone-6 work. Linux and macOS are unaffected.

Fix: resolve the child through the same helper the loader uses (or append the platform suffix), and make a missing child exit the harness with a failure code instead of waiting.

marc-korte:
Closes #504

## The bug

Two things, both of which end the same way.

`src/DuskStudioApp.cpp` resolved the harness child as `exe.getSiblingFile ("dusk-studio-plugin-host")`. On Windows the installed tree ships `dusk-studio-plugin-host.exe`; the loader's `getHostExecutablePath()` already appended the suffix, the harnesses did not.

Both harness blocks were also `#if JUCE_LINUX`, so on Windows `DUSKSTUDIO_RUN_IPC_SELFTEST` and `DUSKSTUDIO_IPC_HOST_TEST` were compiled out and silently ignored, and the app started its GUI instead of the harness.

Confirmed on the Windows 11 VM against the installed 0.13.3 build, before this change:

```
exe: C:\Program Files\Dusk Studio\bin\DuskStudio.exe
siblings matching plugin-host:
  dusk-studio-plugin-host.exe
exists no-suffix : False
exists .exe      : True
started pid 10728, waiting up to 60s
STILL RUNNING AFTER 60s (hang)
--- stdout ---
--- stderr ---
[DuskStudio] parallel strip DSP: 10 worker(s)
```

No harness output, no exit.

## The fix

- `pluginHostExecutableName()` in `src/engine/PluginManager.h` owns the child's file name and its Windows suffix. `getHostExecutablePath()` builds on it, and so do both harnesses, so the two cannot drift again.
- The harness gate widens from `JUCE_LINUX` to `DUSKSTUDIO_HAS_OOP_PLUGINS`, which is every platform that builds the child. `IpcSelfTest.cpp` was already compiled on all three.
- A missing child prints the name it looked for and exits 1 instead of handing the connect a path that can never be spawned.

No behaviour change on the Linux path beyond the new guard.

## Verification

Linux:

- `DUSKSTUDIO_RUN_IPC_SELFTEST=1` under Xvfb, child present: `PASS`, exit 0, 10000 round-trips.
- Same run with the child moved aside: exit 1 in under a second with `FAIL: no dusk-studio-plugin-host next to the app; build it first`. That is the hang this fixes, in its Linux-visible form.
- `cmake --build build -j3` clean, no new warnings.
- `ctest --test-dir build-tests` 981/981 pass.
- `tools/juce-gate.sh` 163 files / 8065 uses, down two from main's 8067. The allowlist ceiling is untouched.

Windows: `windows-tests.yml` builds the `DuskStudio` target on MSVC, so this compiles the widened gate there, and the new Catch2 case asserts the `.exe` name under `_WIN32` in that same job. I did not run the fixed binary on the VM: it has no source tree or dependency set, and standing it up is the full CI recipe. The pre-fix hang above is reproduced from the shipped build.

## Test

`tests/plugin_manager_descriptor.cpp` pins the child's name per platform. README's test-case count goes 1001 to 1002; #538, #540 and #541 bump the same line, so whichever lands last re-bumps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1001 Catch2 test cases across 188 test source files. Linux
+The C++ suite declares 1002 Catch2 test cases across 188 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
@@ -123,7 +123,7 @@ src/
   session/     # Session model + JSON serialisation
   ui/          # MainComponent, ConsoleView, channel/aux/master strips, mastering view
   util/        # Native log storage + CrashHandler signal reports
-tests/         # 1001 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
+tests/         # 1002 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
 packaging/     # .desktop, AppStream, MIME, macOS bundle — for tarball + DMG builds
 DuskStudio.md  # authoritative product spec
 MANUAL.md      # end-user manual (Pandoc-buildable to PDF via docs/build-pdf.sh)
--- a/src/DuskStudioApp.cpp
+++ b/src/DuskStudioApp.cpp
@@ -30,7 +30,7 @@
 #include "session/SessionSerializer.h"
 #include "util/CrashHandler.h"
 #include "util/SingleInstance.h"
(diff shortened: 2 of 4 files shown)


marc-korte:

Issue #533, opened 2026-09-09
[Docs] Quickstart page for the first five minutes, linked from the app and every package


marc-korte:
## Context

MANUAL.md has a "Getting started" section, but it sits inside a manual that renders to a large PDF. A new user who has just installed 1.0 needs a page that fits on one screen and gets them from launch to a bounced file. There is no such page today, and nothing in the app points at one.

## Work

- Write a quickstart page covering exactly the first five minutes: pick an audio device, create a session from a template, arm a track, set an input level, record, play back, bounce the master, and where the files landed.
- Keep it to one page. Link out to MANUAL.md for everything else.
- Link it from the app. The startup screen and the Help surface are the two places a new user will look.
- Include the link in `packaging/README-linux.txt`, the DMG, and the MSI so it is reachable before the app is even open.
- If a screencast is recorded, link it from the same page. The written page is the deliverable; the video is optional.

## Acceptance criteria

- The quickstart page exists and a reader who has never opened Dusk Studio can follow it end to end without opening MANUAL.md.
- The app exposes the link from at least the startup screen.
- The link is present in the Linux tarball README, in the DMG, and in the MSI.
- The link resolves for a user who is not signed in to GitHub, or the page ships inside the package.

marc-korte:
The in-app link is left for Worker A, after #550 lands: the startup surface it should hang off is the new-session path being reworked there, and wiring a link into it now would collide.

The page itself and its links from README and MANUAL are in the PR above. The packaging half, making every package ship the real text instead of the one-line pointer, lands on the #546 lineage, where `packaging/QUICKSTART.txt` and `packaging/contents.txt` actually live.

Issue #320, opened 2026-08-19
Route macOS builds to the dusk-mac-air self-hosted runner


marc-korte:
## Problem

The M3 MacBook Air was set up as a headless macOS build node with a registered GitHub Actions runner, but nothing routes work to it. It carried none of the 0.13.0 release.

Current state:

- Runner `macbook-air-m3` is registered on this repo with labels `self-hosted, macOS, ARM64, dusk-mac-air`, and reports `status=offline`.
- `grep -rn 'dusk-mac-air\|self-hosted' .github/` returns zero hits. No workflow has ever referenced the label.
- `macos-release.yml:34` and `macos-build.yml:27` both pin `runs-on: macos-14`.

So `dusk-studio-0.13.0-macOS-arm64.dmg` was built on a GitHub-hosted `macos-14` runner (10m17s, success). That runner is Apple Silicon, so the shipped arch is correct — this is a cost and capacity question, not a correctness one.

## Why bother

GitHub-hosted macOS minutes bill at 10x the Linux rate. The macOS release job alone is ~10 minutes per tag, and `macos-build.yml` runs on every push and PR. Moving those to hardware already sitting on the network removes the largest recurring Actions cost in the repo.

## Work

1. Bring the runner back online. It is registered but offline — check the launchd agent on the Air, and confirm it stays up across reboot and lid-close on a headless box.
2. Provision the toolchain. The GitHub-hosted image supplies Xcode, cmake and ninja for free; the Air does not. Constraints on that machine: no `sudo`, ever (password-gated), so everything installs under `~`. Tools live in `~/bin` and `~/tools` and must be on the runner service's `PATH`, which is not the interactive shell's `PATH`.
3. Verify the sibling checkouts the build discovers — `../JUCE` (upstream on macOS) and `../plugins` at the pinned `DONOR_REV` — exist on the Air and are at the right revisions.
4. Switch `runs-on` to `[self-hosted, macOS, ARM64, dusk-mac-air]` in `macos-build.yml` first. Leave `macos-release.yml` on `macos-14` until the build workflow has been green on the Air across several PRs.
5. Only then move `macos-release.yml`. A tag build that fails on unavailable hardware costs a re-tag.
6. Decide the fallback. A self-hosted job queues indefinitely when the runner is offline rather than failing fast. Either set a job-level `timeout-minutes` so an asleep Air surfaces quickly, or keep a `macos-14` fallback path for tag builds.

## Constraints

[...]

Issue #75, opened 2026-07-15
New Feature: Built In Plugins


marc-korte:
In an attempt to make Dusk Studio  a one stop shop, add a "Built In DSP" section to the insert modal for Dusk Audio plugins. that way the user does not need to download anything and have all tools like a synth, reverb, delay, etc. built into the DAW.

marc-korte:
Scoping this for 1.0.

**What already exists.** The channel strips, aux buses and master already run Dusk Audio DSP compiled straight into the app: high-pass, four-band EQ, FET and opto compressors, Pultec-style master EQ, bus compressor, tape saturation, multiband compression and the brickwall limiter. None of that is a plugin the user can insert. It is fixed in the signal path.

**What is missing.** The insert slots and the aux slots can only take a third-party CLAP, LV2, VST3 or AU. Out of the box, with no downloads, a user has no reverb, no delay and no synth. That is the gap this issue closes, and it is the difference between a DAW that works on its own and one that needs a shopping trip first.

**Proposed 1.0 scope: five built-in inserts.**

- Reverb
- Delay or tape echo
- A saturation or amp colour unit
- A utility unit (gain, width, polarity, meter)
- One instrument, so a MIDI track makes a sound with nothing installed

The donor repo already carries cores for most of these (`DuskVerb`, `tape-echo`, `DuskAmp`, `multi-synth`). The work is a built-in insert category in the picker, a slot type that resolves to internal DSP rather than a scanned file, session serialization for that slot type, and a small editor per unit.

**Constraints.**

- The DAF and DAF-Widgets repositories are out of bounds for 1.0. Consuming them is already how the native notepad UI is built, and that stays. Do not modify them.
- The donor pin `DONOR_REV` is shared by every workflow. If a built-in unit needs donor sources the pin does not carry, moving the pin is its own change and every workflow site moves together.
- No work here may be justified as reducing framework coupling. That campaign resumes after 1.0.

**Acceptance criteria.**

- A fresh install with no third-party plugins installed exposes a Built-In category in the insert picker with at least three units and no more than five.
- Each unit loads on a channel insert and on an aux slot, processes audio, and reports its latency correctly.
- Each unit's parameters save and restore across a session save and reload.
- Each unit is covered by a Catch2 test asserting at minimum silence in gives silence out and unity settings give unity gain.
- The suite ships in the Linux tarball, the DMG and the MSI, verified by the packaging check in #534.
[...]

PR #546, opened 2026-09-09
build(packaging): define and verify the 1.0 package contents

### LINKED ISSUE #534: [Packaging] Define the 1.0 package contents and verify all three packagers against it
## Context

Three packagers exist and each was grown separately. `scripts/package-tarball.sh` stages the app, the plugin host, desktop and MIME integration files, an install script, `LICENSE` and `LICENSES.txt`. CPack builds the MSI and the DMG from the `install()` rules in CMakeLists.txt. Nothing states what a 1.0 package is supposed to contain, and nothing checks the three against each other.

That matters more once the built-in plugin suite (#75) exists, because a suite that ships on Linux and not on Windows is worse than no suite at all.

## Work

- Write down the 1.0 package contents as one list: the app, `dusk-studio-plugin-host`, the built-in plugin suite from #75, the quickstart link, `LICENSE` and `LICENSES.txt`, and platform integration files where the platform has them.
- Project templates already exist in the app through `src/session/SessionTemplates.h` (Blank, Band, Beats, Singer-Songwriter). Confirm they reach the File menu on all three platforms rather than shipping template files.
- Reconcile the three packagers against that list.
- Add a packaging check that asserts the list, so a missing file fails the release job rather than the user's first launch.

## Acceptance criteria

- A single documented list of package contents exists and each of the three packagers is checked against it in CI.
- Unpacking each of the three artifacts and enumerating its files matches that list, with only platform integration files differing.
- Removing any listed item from a packager makes the release job fail.
- New Session offers all four templates on Linux, macOS and Windows.

marc-korte:
Closes #534.

## The contract

`packaging/contents.txt`: one `<platform><TAB><path>` record per required file, relative to that package's own install root. 11 paths for Linux, 5 each for macOS and Windows. Comments carry the reasoning, including where each install root is and why Windows is different.

Two deliberate absences, both recorded in the file:

- **Project templates.** My own acceptance criteria on #534 asked for template files. There are none: Blank, Band, Beats and Singer-Songwriter ship as code in `src/session/SessionTemplates.h` and reach the user through File -> New. Nothing to verify, so the contract says so rather than leaving a reader to wonder.
- **The built-in plugin suite** has a commented-out slot per platform. #75 uncomments three lines instead of remembering that three packagers each need a new entry, which is exactly the failure mode this issue exists to prevent.

## The checker

`scripts/verify-package-contents.sh <platform> <root>`, read-only, exits non-zero listing **every** missing path so one run reports the whole gap. It fails closed on a contract line it cannot parse or a platform tag it does not know: silently skipping a malformed record would drop whatever that record required, which is the same class of bug as shipping without the file.

Windows matches by file name anywhere under the extraction rather than by path. An MSI is a database and 7z flattens it, so the recorded layout is informational there. Documented in the contract, the script and the guide.

## Wiring

Each of the three packaging jobs runs it after packaging and before staging: Linux untars, macOS mounts the DMG it just built and detaches on exit, Windows extracts the MSI with 7z. Nothing about publication changes and no tag is pushed.

## Verification

- Built a real tarball with `scripts/package-tarball.sh` and checked it: **11 paths OK**.
- Deleted `LICENSES.txt` and the AppStream file from the unpacked tree and re-ran: **exit 1, both named in one run**. The checker is not vacuous.
- `tests/package_contents.sh`, wired in as the `package-contents-checker` ctest case, covers the happy path, a single missing file, two missing files reported together, the Windows name-matching mode, a malformed contract, an unknown platform tag, and that the shipped contract lists something for all three platforms.
- `ctest` 981/981. App builds clean at `-j3`.
- No README test-count change: this adds a ctest shell case, not a Catch2 `TEST_CASE`.

`QUICKSTART.txt` now ships in all three packages as a one-line pointer for #533 to replace, so the contract has a real slot rather than a promise.

MAINTAINER-GUIDE Part 10 gains a paragraph before "Tag assets and acceptance".

### Windows, against a real MSI

`gh release download v0.13.2 -R dusk-audio/dusk-studio-releases`, extracted with `7z x`. The names are mangled exactly as feared:

```
CM_FP_bin.DuskStudio.exe
CM_FP_bin.dusk_studio_plugin_host.exe
CM_FP_bin.libgallium_wgl.dll
CM_FP_bin.opengl32.dll
CM_FP_LICENSE
CM_FP_LICENSES.txt
```

WiX prefixes the component, folds the install directory into the name with a dot, and **rewrites hyphens as underscores**. The bare-final-component matching I first wrote found none of them, so the Windows release job would have failed on the next tag. Matching now lowercases, folds `-` to `_`, and requires the expected name at a `.` or `_` boundary so `LICENSE` cannot be satisfied by `LICENSES.txt`.

Against that real extraction:

```
error: windows package is missing 1 required path(s):
  QUICKSTART.txt
```

Correct: the four paths 0.13.2 contains all match through the mangling, and only `QUICKSTART.txt` is missing because that release predates it. Adding a `CM_FP_QUICKSTART.txt` to the extraction gives `windows package contents OK (5 paths checked)`, and removing the mangled plugin host again names `bin/dusk-studio-plugin-host.exe`. The real name shapes and the LICENSE-vs-LICENSES substitution are now cases in `tests/package_contents.sh`.

### macOS, against a real DMG

Built on the build Mac and packaged with `cpack -G DragNDrop`, then mounted and checked exactly as the workflow step does:

```
--- mounted contents ---
DuskStudio.app
LICENSE
LICENSES.txt
QUICKSTART.txt
--- checker ---
macos package contents OK (5 paths checked)
```

That includes the helper inside `DuskStudio.app/Contents/MacOS/`, which the contract records because the bundle carries it rather than it sitting beside the app.

Reproducing the step also caught a second bug in my own workflow code. The DMG carries a licence agreement, so `hdiutil` blocks for an answer and closes stdin, SIGPIPEing the `yes` feeding it; under `pipefail` the pipeline reports 141 and the step fails on a mount that worked. It now reads `PIPESTATUS` for hdiutil, the same pattern the launch check below it already uses. Two bugs found by running the thing rather than reasoning about it.

### Linux

Real tarball from `scripts/package-tarball.sh`: 11 paths OK, and 1 with `LICENSES.txt` plus the AppStream file deleted, both named in a single run.

## Still unexercised

The `release.yml` steps themselves run only on the next tag or a manual dispatch. The checker and its inputs are proven on all three platforms, but the YAML around them is not. Dispatch once before merge if you want that proof; I have not spent it.

Unrelated finding while using the Mac: `cpack` is not on that machine's PATH, only `cmake`, `ctest` and `ninja` are symlinked into `~/bin`. I used the absolute path rather than changing Marc's environment. It does not affect CI, which uses GitHub-hosted runners, but it would matter if #320 ever routes macOS packaging to the Air.

## Summary by CodeRabbit

- **New Features**
  - Added a Quickstart guide to packaged releases, alongside the full user manual and license files.
  - Standardized the expected contents of Linux, macOS, and Windows packages.

- **Bug Fixes**
  - Release builds now verify packaged files and platform-specific packaging steps before upload, helping prevent incomplete or inconsistent downloads.

- **Documentation**
  - Added maintainer documentation describing package contents and verification requirements.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -673,8 +673,15 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          msi="$(ls *.msi build/*.msi 2>/dev/null | head -1)"
-          [[ -n "$msi" ]] || { echo "::error::no MSI to verify" >&2; exit 1; }
+          # The repo root only, and exactly one: that is where the packaging
+          # step leaves the MSI and what the upload path publishes, so checking
+          # a copy under build/ could verify a file nobody ships.
+          shopt -s nullglob
+          msis=(*.msi)
+          (( ${#msis[@]} == 1 )) || {
+            echo "::error::expected exactly one MSI at the repo root, found ${#msis[@]}" >&2
+            exit 1; }
+          msi="${msis[0]}"
           work="$(mktemp -d)"
           7z x -y -o"$work" "$msi" > /dev/null
           scripts/verify-package-contents.sh windows "$work"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -892,8 +892,18 @@ jobs:
           set -euo pipefail
           dmg="$(ls *.dmg | head -1)"
           mnt="$(mktemp -d)"
+          # The DMG carries a licence agreement, so hdiutil blocks for an
+          # answer, and it closes stdin once it has one. Read hdiutil's own
+          # status rather than the pipeline's, which pipefail would otherwise
+          # take from the SIGPIPEd yes. Same pattern as the launch check below.
+          set +o pipefail
           yes | hdiutil attach -nobrowse -readonly -noverify -noautoopen \
                         -mountpoint "$mnt" "$dmg" > /dev/null
+          mount_rc=${PIPESTATUS[1]}
+          set -o pipefail
+          [[ $mount_rc -eq 0 ]] || {
+              echo "::error::could not mount $dmg to verify its contents" >&2
+              exit 1; }
           trap 'hdiutil detach "$mnt" >/dev/null 2>&1 || true' EXIT
           scripts/verify-package-contents.sh macos "$mnt"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -336,6 +336,15 @@ jobs:
         # integration.
         run: BUILD_DIR=build-linux scripts/package-tarball.sh
 
+      - name: Verify tarball contents
+        # packaging/contents.txt is the contract; a packager that stops shipping
+        # a file fails here rather than on the user's first launch.
+        run: |
+          set -euo pipefail
+          work="$(mktemp -d)"
+          tar -xf dusk-studio-*-Linux-${{ matrix.arch }}.tar.xz -C "$work"
+          root="$(find "$work" -mindepth 1 -maxdepth 1 -type d -print -quit)"
+          scripts/verify-package-contents.sh linux "$root"
 
       - name: Stage Linux payload
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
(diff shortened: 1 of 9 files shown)


marc-korte:
Four findings, verified individually against the branch before acting.

**1. hdiutil and pipefail in Verify DMG contents: already addressed.** The step reads hdiutil's own status through `PIPESTATUS[1]` between `set +o pipefail` and `set -o pipefail`. I hit this exact failure reproducing the step on the build Mac and fixed it in `eb604ab2` before this review landed.

**2. MSI selection: fixed.** Valid. `ls *.msi build/*.msi | head -1` could have verified a copy under `build/` while the upload path publishes `*.msi` from the repo root, so the verified file and the shipped file could differ. It now enumerates the root only with `nullglob` and requires exactly one, erroring with the count otherwise.

**3. packaging/contents.txt plugin-suite lines: skipped.** The built-in suite does not exist until #75 ships. Uncommenting those entries would fail every release between now and then. The commented slot is deliberate, and the comment above it says so: it is there so #75 uncomments three lines rather than having to remember that three packagers each need a new entry.

**4. Empty contract paths: fixed.** Valid, and worse than it reads. The check is `-e "$ROOT/$path"`, so an empty path resolves to `$ROOT/`, which is a directory and always exists. A mistyped record became a check that could not fail. Reproduced it (`linux&lt;TAB&gt;` with an empty second field passed against an empty directory), then rejected it as malformed with the same fail-closed error and exit 2 as the other shape violations. `tests/package_contents.sh` gains a case asserting the contract line number is named.

`tests/package_contents.sh` passes, `shellcheck -S warning` is clean, ctest 981/981.

PR #552, opened 2026-09-09
ci(windows): upload the built app and plugin host as a short-lived artifact

marc-korte:
The Windows job already builds the `DuskStudio` target, then throws the output away. The MSI is tag-driven, so there is no route to a runnable Windows build from a branch, which is what walking the demo path on the test VM needs.

Uploads `build-tests/DuskStudio_artefacts/Release/` (the app, `dusk-studio-plugin-host.exe`, and the runtime files beside them) as `duskstudio-windows-x64`, 3-day retention, `if-no-files-found: error` so a layout change fails loudly instead of producing an empty download. Pinned to the same upload-artifact SHA `release.yml` uses.

The step sits after the build and before the tests, so a red test still leaves a binary to investigate with.

Nothing else in the workflow changes.

### DIFF
--- a/.github/workflows/windows-tests.yml
+++ b/.github/workflows/windows-tests.yml
@@ -144,6 +144,17 @@ jobs:
         shell: bash
         run: cmake --build build-tests --config Release --target DuskStudio dusk-studio-tests -j4
 
+      - name: Upload the built app
+        # Windows has no other route to a runnable build from a branch: the
+        # MSI is tag-driven. Short retention because this is for walking a
+        # branch on a test machine, not for distribution.
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: duskstudio-windows-x64
+          path: build-tests/DuskStudio_artefacts/Release/
+          retention-days: 3
+          if-no-files-found: error
+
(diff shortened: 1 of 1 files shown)


marc-korte:
Note for whoever uses this artifact next, from taking it to the Windows VM.

The binary this step uploads is built without `DUSKSTUDIO_WINDOWS_SOFTWARE_OPENGL_DIR`, so it lacks `DUSKSTUDIO_USE_WINDOWS_SOFTWARE_OPENGL` and never runs the `GALLIUM_DRIVER=llvmpipe` selection in `Main.cpp`. On a machine whose system OpenGL is GDI Generic 1.1 (a VM, or Remote Desktop) it exits immediately with no window and only a console line.

Copying the MSI's `opengl32.dll` and `libgallium_wgl.dll` beside it makes it start, but Mesa is then present and unselected and picks the D3D12 backend on the Basic Render Driver, which hangs the first frame. I lost an hour to that and filed #559 against the product before working out it was the artifact; #559 is closed.

Nothing to change in this PR, which does what it says. Worth knowing that the artifact is a test build, not a stand-in for the MSI, and that walking a branch on a GPU-less machine needs either the Mesa payload alongside or `GALLIUM_DRIVER=llvmpipe` in the environment.

Issue #550, opened 2026-09-09
[Startup] Creating the first session offers no template choice


marc-korte:
Found walking the demo path on a packaged Linux build (#536).

## Steps

1. Launch with no prior sessions. The startup picker opens.
2. Click **NEW**.

## Expected

The chance to start from one of the templates, since this is the moment a first-time user decides what they are making.

## Actual

**NEW** goes straight to "Name your new session" with a path and a name field. No template choice is offered, so a first session is always Blank.

The templates do exist and are reachable, but only from the menu bar once a session is already open: **File -> New from template** with a submenu. A first-time user has no reason to go there, and by the time they find it they have already started from Blank.

Blank / Band / Beats / Singer-Songwriter are part of the 1.0 story and cost the user nothing to skip, so offering them where the first session is created seems like the point of having them.

## Screenshot

The "Name your new session" dialog with path, file list, Name field and Save / Cancel, and separately the File menu showing "New from template" with its submenu arrow.

PR #553, opened 2026-09-09
fix(session): a never-saved session is not dirty just because it autosaved

### LINKED ISSUE #549: [Startup] First launch asks to save changes to a session the user has not touched

## Steps

1. Unpack the 0.13.3 Linux tarball, `./install.sh` into a clean HOME, launch the installed binary. Nothing has ever been created or opened.
2. The startup picker opens on **RECENT**, showing "No recent sessions yet." with Open disabled.
3. Click **NEW**.

## Expected

Go straight to naming the new session. There is nothing to lose.

## Actual

A modal: **"Save changes before starting a new session?"** with the body "Your current session has unsaved changes. If you don't save, those changes are discarded when the new session opens." and Cancel / Don't Save / Save.

The user has not touched anything. The default session behind the picker is being treated as dirty on first run, so the very first click in the application asks them to make a decision about work they have not done and cannot identify.

Answering **Don't Save** proceeds correctly to the naming dialog, so it is one unnecessary step rather than a dead end, but it is an odd first impression and it teaches the user to dismiss save prompts.

Related first-run observation, mentioned here rather than filed separately: the picker opens on RECENT, which for a first-time user is guaranteed to be empty with a disabled Open button. Landing on NEW when there are no recent sessions would put the only useful action under the cursor.

## Screenshot

The save-changes modal centred over the mixer, with the startup picker behind it.

marc-korte:
Closes #549

## Root cause

Reproduced on a clean `HOME` under Xvfb. The bootstrap session directory after 30 seconds of an untouched app:

```
~/Dusk Studio/Untitled/
  audio/
  session.json.autosave
```

No `session.json`. The user has not saved, so there is not one. The autosave heartbeat writes its file into that directory regardless, and `autosaveIsNewerThan` returns true whenever `session.json` is absent, with the comment "autosave is the only state we have". That reading is right for the recovery-offer path at load. It is wrong for the dirty check, where the file it is comparing against was written by us, from the very state being compared.

So from the first autosave tick, an untouched app reads dirty forever, and the first click a new user makes asks them to make a decision about work they have not done.

## Fix

With no `session.json` on disk, the baseline compare (live JSON vs the baseline seeded at construction) is the whole answer. `autosaveIsNewerThan` is untouched, so the recovery path at load still reads a lone autosave as recoverable state.

Genuine edits to a never-saved session still prompt: the baseline is seeded in the constructor, so any divergence is caught.

## Verification

Clean `HOME`, Xvfb, `session.json.autosave` present and no `session.json`, clicking NEW in the startup picker goes straight to "Name your new session". Screenshots before and after the click are in the session scratchpad. Build clean, no new warnings, ctest 981/981, `tools/juce-gate.sh` unchanged at 163 / 8067.

No unit test: the check reads `juce::File` existence and calls into the engine to publish plugin state, so it is not reachable from the session-level test harness. The filesystem repro above is what pins it.

## Not included

#549 also notes that the picker lands on RECENT, which is empty on first run. RECENT is not a tab in the sense the label suggests, it is the only panel the dialog draws, with OPEN and NEW as actions that dismiss it. Changing where the dialog lands is a design decision about the first-run path rather than a bug fix, and #550 is where that path gets its attention.

### DIFF
--- a/src/ui/MainComponent.cpp
+++ b/src/ui/MainComponent.cpp
@@ -3315,8 +3315,23 @@ bool MainComponent::currentSessionDirty()
     // so transient fields don't flip the raw JSON on an otherwise-clean session.
     const auto strippedCurrent = stripVolatileStateForDirtyCompare (SessionSerializer::serialize (session));
     const auto strippedSaved   = stripVolatileStateForDirtyCompare (lastSavedSessionJson);
-    return (! lastSavedSessionJson.isEmpty() && strippedCurrent != strippedSaved)
-         || autosaveIsNewerThan (dir.getChildFile ("session.json"));
+    const bool divergedFromBaseline =
+        ! lastSavedSessionJson.isEmpty() && strippedCurrent != strippedSaved;
+
+    // A session that was never saved has no session.json, only the autosave the
+    // heartbeat wrote from this very state. autosaveIsNewerThan answers "yes"
+    // there because there is nothing for it to be newer than, which on the
+    // bootstrap Untitled session means the first autosave tick makes an
+    // untouched app ask the user to save work they have not done. The baseline
+    // seeded at construction answers the question instead - and if that
+    // baseline is ever missing, the old reading stands rather than reporting an
+    // hour of unsaved work clean.
+    const auto sessionJson = dir.getChildFile ("session.json");
+    if (! sessionJson.existsAsFile())
+        return lastSavedSessionJson.isEmpty() ? autosaveIsNewerThan (sessionJson)
+                                              : divergedFromBaseline;
+
+    return divergedFromBaseline || autosaveIsNewerThan (sessionJson);
 }
 
 void MainComponent::requestQuit()


marc-korte:
Taking the objection first, because it is the right one to raise: with the early return, a never-saved session that never sets a baseline would read clean forever, and that is worse than the bug.

## (a) The baseline is set at creation

Two sites, both predating this change.

`src/ui/MainComponent.cpp:1100-1101`, the constructor, which is what the bootstrap Untitled session gets:

```cpp
    if (lastSavedSessionJson.isEmpty())
        lastSavedSessionJson = SessionSerializer::serialize (session);
```

`src/ui/MainComponent.cpp:3252-3258`, `setLastSavedSessionJson`, called on every save and at the tail of `finishLoadingSessionFrom`:

```cpp
void MainComponent::setLastSavedSessionJson (const juce::String& json)
{
    lastSavedSessionJson = json;
```

The seed placement is deliberate and documented at `src/ui/MainComponent.cpp:3874`: it happens at the tail of the load, after `consumePluginStateAfterLoad` has filled in plugin and transport state, so the snapshot matches what the next compare will serialize.

There is also a structural reason File > New is not exposed: `createNewSessionAt` writes `session.json` before loading it, so from that point on `session.json` exists and the early return never fires. That path keeps the old logic byte for byte. The early return only ever applies to the bootstrap Untitled session, which is the one the constructor seeds.

## (b) Done anyway

Argument is not proof, so the branch now does what you asked for in (b). If the baseline is ever empty, the old reading stands:

```cpp
    const auto sessionJson = dir.getChildFile ("session.json");
    if (! sessionJson.existsAsFile())
        return lastSavedSessionJson.isEmpty() ? autosaveIsNewerThan (sessionJson)
                                              : divergedFromBaseline;
```

Two lines, and the failure mode you described is now unreachable by construction rather than by argument. Force-pushed.

## The walk, honestly

I could not complete a scripted quit-with-an-edit walk under Xvfb, and I am not going to claim I did.

What worked: fresh HOME, launch, `Escape` to dismiss the picker, four taps of `B` moved the tempo from 120 to 128.66, which is a real serialized edit. What did not: driving the quit. `Ctrl+Q` does nothing on Linux either. The File menu opens under xdotool and Quit highlights correctly at its coordinates, but the click landed on New session the first time and reopened the startup picker the second, and there is no window manager on that display so there is no titlebar X to use instead. Rather than keep fighting the harness I made the code safe by construction, which is the stronger answer to the question you asked.

What is verified on the branch: fresh HOME with `session.json.autosave` present and no `session.json`, clicking NEW goes straight to "Name your new session" with no save prompt. Build clean, no new warnings, ctest 981/981, gate unchanged at 163 / 8067.

One thing worth noting for whoever picks up the quit path: `Ctrl+Q` not quitting on Linux may be the same defect as #558, which I filed for Windows during the walk. I have not chased it here.

marc-korte:

Issue #548, opened 2026-09-09
[Audio] First launch selects an output device but leaves the input as (None)


marc-korte:
Found walking the demo path on a packaged Linux build (#536). Trigger for #547.

## Steps

1. Unpack the 0.13.3 Linux tarball, `./install.sh` into a clean HOME, launch the installed binary. No prior configuration of any kind.
2. Settings -> Settings...

## Expected

An input device selected, on a machine that has several.

## Actual

```
Audio backend:  ALSA
Output device:  HDA Intel PCH, ALC700 Analog
Input device:   (None)
```

An output device was chosen automatically; the input was left unset. This machine has capture devices, and the self-test enumerates them:

```
in:  HDA Intel PCH, ALC700 Analog
in:  UMC1820, USB Audio
in:  HD Pro Webcam C920, USB Audio
```

So a device that could have been selected was available and was not.

The startup log also shows the backend falling back:

```
[Dusk Studio/AudioEngine] device-manager init reported: PipeWire delivered no usable quantum (0)
[Dusk Studio/ALSA] opened "..." out=2ch in=0ch
```

Whether the PipeWire fallback is what leaves the input unset is worth checking; if it is, every user whose PipeWire probe fails lands in the same state.

## Why it matters

On its own this is a settings default. Combined with #547, where recording in this state silently does nothing, it is the first thing that breaks the demo path. Picking a sane default input, or the first input on the chosen device, would fix the common case.

## Screenshot

Settings panel, Audio section, showing the three rows quoted above.

Issue #536, opened 2026-09-09
[UX] Walk the demo path on packaged builds and file what it snags on


marc-korte:
## Context

The 1.0 pitch is that someone downloads Dusk Studio, opens it, records something and bounces it, inside five minutes, without reading anything. Individual bugs on that path have been fixed one at a time (#109 latency offset, #193 arm and monitoring, #87 list scrolling). Nobody has walked the whole path on a shipped build and written down every place it snags.

## Work

Walk the path on a packaged build for each of Linux, macOS and Windows, under Xvfb on Linux, and record every snag:

1. First launch with no configuration. Does the audio device pick sane defaults, or does the user land in a settings panel they do not understand?
2. New session from a template. Is the destination folder obvious? Is the session saved before recording starts?
3. Arm a track and set a level. Is there enough feedback to know the signal is arriving and not clipping?
4. Record, stop, play back. Is the take where it is expected?
5. Overdub a second track against the first.
6. Bounce the master. Is the output path obvious and is the file where the dialog said?

For each snag record: what happened, what a first-time user would expect, and how large the fix is. File the ones worth fixing as separate issues labelled `1.0`, linked here. Close this issue when the walk is documented and the follow-ups are filed.

## Acceptance criteria

- The walk is completed and written up for all three platforms on a packaged build, not a build tree.
- Every snag is either filed as a linked issue or explicitly recorded as accepted for 1.0 with a reason.
- The written path matches the quickstart page from the quickstart issue, or one of the two is corrected so they agree.

marc-korte:
## Linux leg: walk log

Packaged 0.13.3 tarball built from `origin/main` (a059c83) with `scripts/package-tarball.sh`, unpacked, and installed with `./install.sh` into a scratch `HOME` so nothing touched the real one. Walked on a private Xvfb display (`:77`), `WAYLAND_DISPLAY` unset, never the live session. Driven with xdotool, screenshotted at every step; the images are in the session scratchpad, not the repo.

| # | Step | Result |
|---|---|---|
| 1 | Unpack, read README-linux.txt | pass. Says plainly that the program runs in place and what install.sh adds. |
| 2 | `./install.sh` into a clean HOME | pass. Installs to `~/.local`, prints the launcher path, warns that `~/.local/bin` is not on PATH. |
| 3 | First launch | pass, with two snags. App starts, opens a device, shows the startup picker. Picker opens on RECENT, which is guaranteed empty on first run, with Open disabled. |
[...]

Issue #501, opened 2026-09-05
[Aux] Inline non-modal editor status in the aux slot area


marc-korte:
Follow-up to #459 and the headless-plugin regression it caused.

Aux lanes attach native editors automatically in rebuildSlots, so a modal alert is the wrong surface: it fires without a user action and, for DSP-only plugins, on every session load. The channel strip only alerts from a click. The aux slot area should instead show an inline, non-modal status where the editor would be: "No editor" when the plugin has none, or the format, plugin name and reason when the attach failed. That also needs a has-editor query on the native instances so the two cases can be told apart.

marc-korte:
Re-triaging this to post-1.0. The bug half is already fixed; what is left is additive UI work.

The unprompted modal is gone. #459's fix wraps every `rebuildSlots` attach in `attachNativePluginEditor` (`src/ui/AuxLaneComponent.cpp:38-58`), which calls `enforceNativeEditorAttachPolicy` with a real log callback and an empty alert callback, for the reason recorded in the comment there: aux lanes attach from `rebuildSlots`, which runs from the constructor for hidden lanes too, so a failure has no user-initiated open to answer and no on-screen parent, and an alert would push an invisible entry onto the modal stack that gates native-editor reopen everywhere else.

The `showDuskAlert` calls that remain in the file (`:1194`, `:1253`, `:1307`, `:1355`) are in `loadNativeClapForSlot` / `Lv2` / `Vst3` / `Au`, whose only callers are the picker callbacks in `openPickerForSlot` (`:921`, `:929`, `:937`). Those are click-initiated, which is the same contract the channel strip has. Session restore does not reach them.

What is still missing:

- A has-editor query on `INativeInstance` (there is none today) with CLAP, LV2 and VST3 implementations, so "the plug-in has no editor" can be told from "the attach failed".
- Inline, non-modal status text in the aux slot editor area carrying either "No editor" or the format, plug-in name and reason.

Both are worth doing, but neither is on the first-run path, and the aux editor area is rewritten by the GUI port (#307, #310). Deferring so the inline surface is built once, against the ported view, rather than twice.

PR #554, opened 2026-09-09
fix(record): refuse to arm an audio track with no capture channels

### LINKED ISSUE #547: [Record] Recording with no input device produces nothing and reports nothing

## Steps

1. Unpack the 0.13.3 Linux tarball, `./install.sh` into a clean HOME, launch the installed binary.
2. Create a session from the startup picker.
3. Click **ARM** on track 1. Click **IN**.
4. Press record. Watch the transport run. Press stop.

## Expected

Either a recorded region on track 1, or a clear statement that there is no input to record from.

## Actual

Nothing. No region on the track, no file in the session's `audio/` directory, no message anywhere. The session's `tracks[0].regions` is still empty. The transport ran and the playhead moved the whole time, and both ARM and IN stayed lit as though it were working.

The cause is visible only in Settings: **Input device is `(None)`** (filed separately). The device opened output-only, which the log confirms:

```
[Dusk Studio/ALSA] opened "HDA Intel PCH, ALC700 Analog" rate=48000 period=1024 periods=3 bits=32 int out=2ch in=0ch
```

## Why this is a blocker

This is the core action of the application and the middle step of the first five minutes. A first-time user arms, records, stops, and has nothing, with no indication of what went wrong or where to look. Nothing on the recording page mentions the input device, so there is no thread to pull.

The fix does not have to be automatic device selection. Refusing to arm, or refusing to record, or a single line on the strip saying there is no input, would each be enough to turn a silent dead end into a solvable problem.

## Screenshot

Track 1 header with IN lit cyan and ARM lit red, next to tracks 2 and 3 in their default state; transport at 3.1.030 with the record button lit, playhead advanced.

marc-korte:
Closes #547. Found walking the demo path (#536); this is the blocker in the middle of it.

## What was wrong

Arm an audio track, press record, watch the transport roll, press stop, and nothing was captured. No region, no file in the session's `audio/`, no message. ARM and IN stayed lit throughout. The device had opened with zero input channels and nothing in the record path noticed.

## The gate

`AudioEngine::audioDeviceAboutToStart` publishes the device's capture width to the session, right beside the existing zero-**output** warning that this is the silent counterpart of. `Session::setTrackArmed` then refuses to arm an audio track while that width is exactly zero, in the same place the frozen-track refusal already lives, so every arm path is covered at once: the strip button, the selected-track shortcut, MCU, and MIDI bindings.

MIDI tracks capture from a MIDI input and are not gated.

**Zero is deliberately distinct from unknown.** The field starts at `kCaptureWidthUnknown` rather than 0. My first version defaulted to zero and broke two existing tests that arm tracks with no device: refusing to arm because no device has reported yet is a different claim from refusing because the device has no inputs, and only the second is true. Rather than adjust those tests I fixed the conflation, which is also what makes headless and pre-device callers behave.

A device change that takes the inputs away disarms the audio tracks it can no longer feed, so ARM is never lit over a dead input.

## The message

Inline in the transport bar, in the gap the controls leave:

No input notice

> No input device. Choose one in Settings > Audio.

Non-modal on purpose. The same condition arises when a device change removes the inputs, where there is no click to answer and a modal would be the #501 mistake. The strip's ARM button now follows the session rather than its own toggle, so a refusal cannot leave it lit, and it stays silent instead of stacking an alert on the same click the bar is already explaining.

## Coverage

Seven Catch2 cases over the gate, no device needed: audio refused at zero, audio armed with channels, MIDI armed regardless, disarm never gated, a device change disarming audio and sparing MIDI, the frozen refusal still standing, and arming allowed while the width is unknown.

## Coupling

Gate reports 163 files / 8059 uses, **down** from 8067. The notice and the ARM rollback cost a few names; hoisting the duplicated colour literals in `TransportBar.cpp` and `ChannelStripComponent.cpp` more than paid for them, and `ChannelStripComponent.cpp` drops 691 to 683 in the allowlist. No file rose and no file was added.

## Screenshot

Captured through the Xvfb harness and inspected. It took two passes: the first drew the notice in the bar's bottom strip, where it would have overlapped the controls, and did not render at all because `settle()` sleeps rather than pumping so the bar's timer never fired. Hence `refreshInputNotice()`, which the timer and the harness both drive, and the centred placement above. Only the new figure is committed; the harness rewrites every PNG with encoding noise and that churn is reverted.

## Verification

`ctest` 987/987. `tools/juce-gate.sh` green. `scripts/run-selftest-xvfb.sh` exit 0, 15/15. App builds clean at `-j3`. MANUAL.md documents the refusal and the message in the ARM section. README test count 1001 -> 1008.

#548, the reason the input was never selected in the first place, is next and separate.


marc-korte:

PR #557, opened 2026-09-09
feat(release): smoke-test a published artifact

marc-korte:
Closes #532. Stacked on #546, whose `verify-package-contents.sh` is one of the checks; review that first.

Not wired into `release.yml`, per the brief. That is the follow-up once the signing steps exist.

## The checks

`scripts/release-smoke-test.sh <linux|macos> <artifact> [version]` and `scripts/release-smoke-test.ps1`. Each unpacks or mounts into a scratch directory removed on exit; nothing is installed and no system location is touched.

1. **Unpack / mount.**
2. **Package contents**, through #546's checker.
3. **`--version`**, required to match the version in the artifact's own file name, so a mismatch between the name and the binary is itself a failure.
4. **Headless self-test** against the packaged binary rather than the build tree, which is the point of the script.
5. **Plugin scan terminates.**

One PASS or FAIL line per leg. Every leg runs even after one fails, so one invocation reports the whole picture.

## Two things the runs changed

**The scan leg started as a second `timeout`-bounded run and both parts were wrong.** macOS ships no GNU `timeout`, so the leg failed outright on the Mac; and re-running the self-test to prove the scan terminates does the same expensive work twice. It is now one bounded run yielding both verdicts, bounded by a wait loop that works on either platform. A check that cannot run on one of the two platforms is not much of a check.

**Linux launches under a private Xvfb display** with `WAYLAND_DISPLAY` cleared, never a live session.

## Exercised, not assumed

Linux, against a tarball I built:

```
PASS  unpack: dusk-studio-0.13.3-Linux-x86_64.tar.xz
PASS  package contents
PASS  --version reports 0.13.3
PASS  plugin scan terminates
PASS  headless self-test
release smoke test: all checks passed (linux, 0.13.3)
```

macOS, on the build Mac against a DMG built there with no sudo:

```
PASS  mount: dusk-studio-0.13.3-macOS-arm64.dmg
PASS  package contents
PASS  --version reports 0.13.3
PASS  plugin scan terminates
PASS  headless self-test
release smoke test: all checks passed (macos, 0.13.3)
```

**Proven non-vacuous**, twice. A wrong expected version gives `FAIL --version reported "Dusk Studio 0.13.3"` and exit 1, with the other legs still running. A tarball rebuilt without `LICENSES.txt` gives `FAIL package contents: ... missing 1 required path(s): LICENSES.txt`.

## Windows

[...]

PR #560, opened 2026-09-09
feat(release): sign SHA256SUMS

### LINKED ISSUE #531: [Release] Publish a signed SHA256SUMS for release artifacts
## Context

The release publisher already creates a `SHA256SUMS` asset covering all five payloads, and `scripts/verify-release-assets.sh` checks the six-asset contract. What is missing is a way for a user to prove the checksum file itself came from us. A checksum published next to the artifact only defends against a corrupt download, not against a replaced release.

Linux has no equivalent of Authenticode or Gatekeeper, so the tarball itself does not need a signature. A signed `SHA256SUMS` is the normal answer.

## Work

- Create a release signing key. Store the private key and its passphrase as repository secrets.
- In the publisher job, after `SHA256SUMS` is created, produce a detached signature and publish it as a seventh asset (`SHA256SUMS.asc`).
- Publish the public key where a user can reach it, and document the verify command in `packaging/README-linux.txt` and the release notes.
- Teach `scripts/verify-release-assets.sh` about the new asset so the six-asset check becomes a seven-asset check rather than failing.

## Acceptance criteria

- A `v*` tag publishes `SHA256SUMS.asc` alongside the existing six assets.
- `gpg --verify SHA256SUMS.asc SHA256SUMS` succeeds against the published public key on a machine that only imported that key.
- `sha256sum --check SHA256SUMS` passes against both downloaded Linux tarballs, the DMG and the MSI.
- `scripts/verify-release-assets.sh vX.Y.Z` exits 0 for a complete release and nonzero when the signature asset is missing.
- The verify steps are documented in `packaging/README-linux.txt`.

marc-korte:
Closes #531. Stacked on #557, which is stacked on #546; review in that order.

## Tool choice: GPG, not minisign

The brief said minisign if it installs from stock apt. I could not verify that from this box (openSUSE, no apt), and guessing wrong means a release job that fails on its first real run. GPG needs no verification: it is already on the ubuntu-22.04 runner, so there is no package to install and no repository question at all. It is also what a user checking a download is far likelier to already have.

## What the publisher does

After `SHA256SUMS` is built and before anything is uploaded: import the key from `RELEASE_SIGNING_KEY` into a keyring under `RUNNER_TEMP`, sign detached and armored to `SHA256SUMS.asc`, shred the passphrase file.

- **A `v*` tag with either secret missing fails the job** before publishing. A release cannot go out unsigned.
- **A `workflow_dispatch` run skips signing with a notice.** It publishes nothing, so there is nothing to protect and no secret worth spending.
- **The signature is verified in the same step that makes it**, against the committed public key. One the key cannot check is worse than none, because it looks like proof.

`packaging/release-signing.pub` is a placeholder with verify instructions until Marc generates the pair. While it holds no key the step warns rather than blocks; the moment it does, a mismatch fails the release. That is what lets this merge before the key exists.

## Exercised end to end with a throwaway key

Generated an ed25519 pair, exported the secret base64 exactly as the secret would be stored, and ran the workflow's own commands:

```
key=B3222D4D4DE1CF38
SIGNED:
-----BEGIN PGP SIGNATURE-----

gpg: Good signature from "Dusk Audio Release Signing (THROWAWAY) <releases@example.invalid>"
```

Then appended a line to `SHA256SUMS`:

```
gpg: BAD signature from "Dusk Audio Release Signing (THROWAWAY) <releases@example.invalid>"
```

And through the smoke test, all three states:

```
SKIP  checksum signature: no release key in packaging/release-signing.pub yet   <- placeholder
PASS  checksum signature                                                        <- key present
PASS  checksum matches the artifact
FAIL  checksum signature does not verify against packaging/release-signing.pub  <- tampered
```

The throwaway pair and its keyring are deleted; nothing was left in any keyring.

## The release contract moved from six assets to seven

`tests/release_mechanics.sh` is a genuine ratchet and it caught every place the number lives, one at a time: the verifier's expected-asset count, `bump-version.sh`'s printed step, the fan-in wording, and a guard asserting no `SHA256SUMS.` suffix ever appears in the workflow. That last one existed to stop per-job checksum fragments returning; I narrowed it to permit exactly `.asc` rather than deleting it, so it still catches what it was written for. Each update is deliberate and visible in the diff, which is the point of that test.

## Verification

`ctest` 980/980 including `release-mechanics-contract`. YAML parses. MAINTAINER-GUIDE Part 10 gains a "Release signing key" section with the exact generation and `gh secret set` commands, and the acceptance asset list moves to seven.

No tag pushed, no dispatch triggered. 182 insertions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -981,7 +981,7 @@ jobs:
           retention-days: 1
 
   publish:
-    name: Assemble + publish six release assets
+    name: Assemble + publish the release assets
     needs: [linux, macos, windows, manual]
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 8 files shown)


marc-korte:

PR #561, opened 2026-09-09
docs: add the quickstart page

marc-korte:
Refs #533. Content half only, from `origin/main`, not stacked.

## The page

`QUICKSTART.md`: one screen, download to mixed-down track. Install and the first-launch bypass each OS currently asks for while builds are unsigned, picking a device, a session from a template, arm and level and record, an overdub, a mix, a bounce, where the files landed, and the four things that go wrong most.

Two details come straight out of the demo-path walk on #536:

- **The input device gets its own sentence**, because "check the input even if the output looks right, if it says (None) you can play back but nothing will record" is exactly the wall I hit walking a packaged build.
- **The overdub step says to turn IN off on the recorded track**, which is the next thing that bites once recording works at all.

Illustrated with `qg-01` through `qg-07`, all of which already exist under `docs/images/`. Nothing new was captured, per the brief. Every referenced figure verified present.

Linked from README (near the manual link, where someone arriving at the repo will see it) and from the head of MANUAL's Getting started chapter.

## What is not here, and why

**The packaging swap.** The brief asks me to replace the one-line `packaging/QUICKSTART.txt` pointer so every package ships the real text, and keep `packaging/contents.txt` green. **Neither of those files exists on `origin/main`**: both arrived in #546, which is unmerged. So the swap cannot be done from an unstacked branch.

Rather than stack this against the instruction or drop the requirement quietly, I have left it out here and will land it as one small commit on `build/534-package-contents`, which is where those files live: point the packagers at `QUICKSTART.md` and update the contract entry from `QUICKSTART.txt` to `QUICKSTART.md`. One source of truth, no plain-text copy to drift. Say the word if you would rather it went elsewhere.

**The in-app link** stays with Worker A after #550, as instructed. Noted on #533.

## Verification

Every image reference resolves. `release-mechanics-contract` still passes. Docs only, no source changes. 120 lines of new page plus two link lines.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -317,6 +317,10 @@ Assign a strip to one of eight fader groups (right-click the strip → **Fader g
 
 # Getting started
 
+If you have just downloaded Dusk Studio and want to be recording rather than
+reading, [QUICKSTART.md](QUICKSTART.md) is the five-minute version of this
+chapter. Everything below is the long form.
+
 ## System requirements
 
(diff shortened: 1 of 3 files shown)


marc-korte:

Issue #529, opened 2026-09-09
[Release] Sign and notarize the macOS DMG with a Developer ID certificate


marc-korte:
## Context

`release.yml` builds the macOS artifact in the job named "Build + package unsigned DMG (macOS arm64 Release)". CMake ad-hoc signs the `.app` with identity `-`, which is enough to launch locally but is not a Developer ID signature. The DMG is not notarized and not stapled, so Gatekeeper on a clean Mac refuses it and the user has to right-click and choose Open. That is acceptable for a beta. It is not acceptable for a 1.0 that we describe as ready to run.

## Work

- Add Apple signing to the macOS release job: sign the `.app` (deep, hardened runtime, timestamped) with a Developer ID Application certificate, then sign the DMG.
- Submit the DMG to `notarytool`, wait for acceptance, and staple the ticket.
- Store the credentials as repository secrets. Required: the Developer ID Application certificate as a base64 `.p12` plus its password, the Apple Team ID, and either an App Store Connect API key (issuer ID, key ID, `.p8`) or an Apple ID with an app-specific password. The API key is preferred because it does not expire on a password change.
- Fail the job if signing or notarization fails. Never publish an unsigned DMG under a `v*` tag.

## Acceptance criteria

- `codesign --verify --deep --strict --verbose=2 DuskStudio.app` exits 0 and reports a Developer ID Application authority.
- `spctl --assess --type execute --verbose DuskStudio.app` reports `accepted` with `source=Notarized Developer ID`.
- `xcrun stapler validate` on the DMG exits 0.
- A clean macOS machine that has never seen the app opens the DMG and launches the app by double click, with no right-click Open and no Gatekeeper prompt.
- The release job fails, and publishes nothing, when the signing secrets are absent or the notarization submission is rejected.

marc-korte:
Deferred: 1.0 ships the macOS DMG ad-hoc signed, without Developer ID signing or notarization. Apple Developer Program enrollment is postponed until the project has revenue to justify it; reopen this issue then.

Until then, the macOS first-open steps in QUICKSTART.md need to cover the Gatekeeper bypass for an ad-hoc signed app on current macOS (tracked in #589).

PR #562, opened 2026-09-09
ci(release): sign and notarize the macOS DMG

marc-korte:
Refs #529, deliberately not Closes: this closes when a signed build is verified end to end, which needs credentials that do not exist yet.

Stacked on #560, which is stacked on #557 and #546.

## What the job does on a tag

Imports the Developer ID certificate into a keychain of its own, unlocked for the job and deleted after it whatever the outcome. The login keychain is never touched.

- **Signs inside out**: the plugin host first, then the bundle. Signing the bundle first would not cover a nested binary signed after it.
- `--options runtime` and `--timestamp`, with the entitlements file the build already produces.
- Packages the DMG, signs that too.
- `notarytool submit --wait`, then `stapler staple` and `stapler validate`. Stapling matters because without the ticket a first launch has to reach Apple, and a user with no network is refused.
- **Acceptance is `spctl --assess`**, required to report `accepted` **and** `source=Notarized Developer ID`. That is what the downloader's machine decides, rather than what our own signing thinks it did.

A `v*` tag with any of the six secrets missing fails before publication. A `workflow_dispatch` run keeps the ad-hoc signature and says so, since it publishes nothing.

## What I verified, and what I did not

**Verified on the build Mac**, against a real built bundle with the ad-hoc identity, so the invocation shape is not guesswork:

```
entitlements present: yes
helper present: yes
helper codesign rc=0
bundle codesign rc=0
DuskStudio.app: valid on disk
DuskStudio.app: satisfies its Designated Requirement
Identifier=audio.dusk.studio
CodeDirectory v=20500 ... flags=0x10002(adhoc,runtime)
```

That confirms the entitlements and helper paths the workflow uses are the real ones, that `--options runtime` genuinely applies the hardened runtime flag, and that the inside-out order passes a deep strict verify with the nested helper validated.

`xcrun notarytool` and `xcrun stapler` both exist on that machine.

**Not verified, and cannot be until Marc adds credentials**: the real Developer ID identity, the secure timestamp server, the notary submission, stapling, and the `spctl` assessment. The keychain import and `set-key-partition-list` dance is written from the documented behaviour, not from a run. The first tag after the secrets land is the real test, and if something is wrong there it will be in that region.

That is why this is Refs and not Closes.

## Credentials

MAINTAINER-GUIDE gains a "macOS signing and notarization" section: Developer Program membership, exporting the Developer ID Application certificate as a `.p12`, creating an App Store Connect API key, and the six `gh secret set` commands. It notes that Developer ID Application is the right certificate type, that Mac App Distribution will not do, and that the `.p8` cannot be re-downloaded.

An API key rather than an Apple ID and app-specific password, because the key does not expire when a password changes and carries no second factor.

## Verification

YAML parses. `release-mechanics-contract` passes. No source changes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -878,9 +878,94 @@ jobs:
           cmake --build build-tests --target dusk-studio-tests -j5
           ctest --test-dir build-tests --output-on-failure
 
-      - name: Package unsigned DMG
-        # CMake already ad-hoc-signed the .app via identity "-"; just wrap
-        # it in a DMG. Not notarized, so users right-click -> Open once.
+      - name: Import the Developer ID certificate
+        # A tag must not ship unsigned: Gatekeeper refuses an unnotarized DMG
+        # and the user is told the app is damaged. A dispatch publishes nothing,
+        # so it keeps the ad-hoc signature CMake already applied and says so.
+        id: signing
+        env:
+          MACOS_CERT_P12_BASE64: ${{ secrets.MACOS_CERT_P12_BASE64 }}
+          MACOS_CERT_PASSWORD: ${{ secrets.MACOS_CERT_PASSWORD }}
+          MACOS_TEAM_ID: ${{ secrets.MACOS_TEAM_ID }}
+          NOTARY_KEY_ID: ${{ secrets.NOTARY_KEY_ID }}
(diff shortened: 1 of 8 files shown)


marc-korte:

Issue #530, opened 2026-09-09
[Release] Sign the Windows MSI and its executables with Authenticode


marc-korte:
## Context

`release.yml` packages the Windows artifact in the step "Package MSI via CPack WIX (unsigned)". The comment there says "No signing args: the MSI ships unsigned by design". An unsigned MSI raises a SmartScreen "Windows protected your PC" warning on every download, and the user has to click through More info and Run anyway. For 1.0 that is a download that most people abandon.

## Work

- Obtain a code-signing certificate. An OV certificate still accrues SmartScreen reputation slowly; an EV certificate carries reputation immediately. Note that since June 2023 both are issued on hardware tokens or through a cloud signing service, so the CI story is either Azure Trusted Signing, a cloud HSM with `signtool` via a CSP, or a self-hosted Windows runner with the token attached.
- Sign both `DuskStudio.exe` and `dusk-studio-plugin-host.exe` before CPack runs, then sign the produced MSI. Signing only the MSI leaves the installed binaries unsigned.
- Timestamp every signature with an RFC 3161 timestamp server so signatures outlive the certificate.
- Store the credentials as repository secrets and fail the job when they are missing on a `v*` tag.

## Acceptance criteria

- `signtool verify /pa /v` on the MSI and on both installed `.exe` files exits 0 and names Dusk Audio as the subject.
- Every verified signature carries a countersignature timestamp.
- Downloading the MSI in Edge on a clean Windows 11 machine and running it produces no "Windows protected your PC" SmartScreen block. Record the observed result; with an OV certificate a first-run warning may persist until reputation builds, and that outcome must be written into the release notes.
- The release job fails, and publishes nothing, when signing fails or the secrets are absent.

marc-korte:
Decision: no paid code-signing certificate before the software earns one, matching #529. The Authenticode path in release.yml stays for when secrets exist; #620 lets a tag ship the unsigned MSI with the documented SmartScreen steps. Free option to revisit for 1.1: SignPath Foundation's OSS program.

PR #563, opened 2026-09-09
ci(release): sign the Windows MSI and its executables

marc-korte:
Refs #530, not Closes: it closes when a signed MSI is verified end to end.

Stacked on #562, so on the #560 / #557 / #546 lineage. Review in that order.

## Order matters

Both executables are signed **before** the MSI is built, then the MSI itself. Signing only the installer would leave `DuskStudio.exe` and `dusk-studio-plugin-host.exe` unsigned once laid down, and those are what the user actually runs. Every signature carries `/tr http://timestamp.digicert.com /td SHA256`, so it outlives the certificate.

## Two credential shapes, and why

Since June 2023 both OV and EV code-signing certificates are issued on hardware tokens or through a cloud service. A `.pfx` you can hand a hosted runner is no longer something a CA will sell, so a PFX-only design would be scaffolding for a route that cannot be taken.

- **Azure Trusted Signing** is the route to plan around: Microsoft holds the key, the runner authenticates to it, no token and no secret certificate. Selected when all six of its secrets are present.
- **A PFX** is still supported for an existing certificate, an internal CA, or a test certificate, and takes precedence when set.
- **Neither complete on a `v*` tag fails the job** before publication. A dispatch skips with a notice.

Worth knowing for the release notes: EV carries SmartScreen reputation immediately, OV and Trusted Signing accrue it as downloads accumulate, so the first users of a fresh certificate may still see the warning.

## Acceptance

`signtool verify /pa /v` on all three files, plus a check that each carries a countersignature timestamp. A signature with no countersignature verifies today and stops verifying when the certificate expires, which is the failure worth catching early.

## Hygiene

The PFX is written under `RUNNER_TEMP` and removed in a `finally` block, so a signing run that throws leaves no certificate on disk. Same discipline as the keychain deletion in #562.

## What is verified: nothing

There is no Windows machine in reach from here and no credentials to exercise it with. `signtool` ships with the Windows SDK on `windows-2022`, which the job already relies on for the WiX toolset, but I have not run any of this. **The whole path is unverified until credentials exist**, and the first tag after that is the real test. If something is wrong it will be in the credential handling or the `signtool` invocation, since those are the parts I could not run.

That is why this is Refs and not Closes.

## Documentation

MAINTAINER-GUIDE gains a "Windows code signing" section: the token-versus-cloud problem, OV against EV and what each means for SmartScreen, Azure Trusted Signing as the recommended route with the app-registration role it needs, and the `gh secret set` commands for both shapes.

## Verification

YAML parses. `release-mechanics-contract` passes. No source changes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -657,15 +657,159 @@ jobs:
           $wixBin                     | Out-File -FilePath $env:GITHUB_PATH -Append -Encoding utf8
           Write-Host "WiX ready: $wixBin"
 
-      - name: Package MSI via CPack WIX (unsigned)
+      - name: Decide the Windows signing route
+        # Two credential shapes. A PFX is the simple one and works with any
+        # certificate you hold; Azure Trusted Signing is the CI-friendly one,
+        # because since June 2023 OV and EV certificates are issued on hardware
+        # tokens that a hosted runner cannot reach. Either is enough; neither on
+        # a tag is not.
+        id: winsign
+        shell: bash
+        env:
+          WINDOWS_CERT_PFX_BASE64: ${{ secrets.WINDOWS_CERT_PFX_BASE64 }}
+          WINDOWS_CERT_PASSWORD: ${{ secrets.WINDOWS_CERT_PASSWORD }}
+          AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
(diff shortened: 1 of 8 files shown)


marc-korte:
Pinned to `azure/trusted-signing-action@c7ab2a863ab5f9a846ddb8265964877ef296ee82 # v2.0.0` in both uses.

Pinning caught a real error: **there is no `v0` tag on that action.** The tags are v1.x and v2.x, so `@v0` would have failed to resolve on the first run with credentials, which is exactly the run nobody wants to debug. Checked v2's `action.yml` before pinning; every input name used here is valid in it.

marc-korte:

PR #564, opened 2026-09-09
feat(startup): offer the templates where the first session is made

### LINKED ISSUE #550: [Startup] Creating the first session offers no template choice

marc-korte:
Closes #550

## The gap

Blank, Band, Beats and Singer-Songwriter existed, but only behind **File > New from template**, which a first-time user has no reason to open. NEW in the startup picker went straight to naming, so the first session anyone makes was always Blank, which is the one moment the choice is worth anything.

## The change

NEW opens a panel in the picker listing the starting points. Picking one dismisses, then the naming dialog opens as before, and the choice travels with the action.

The stamping happens on the fresh `Session` before it is serialized, so the session lands on disk the way the user asked for it and the normal load path rebuilds the UI from the file. That is a different route from the menu path, which applies a template to the live session and rebuilds the console view by hand; nothing there changes.

`makeStartupView` takes the labels from the host and hands back an index, so `SessionTemplate` stays the single source of what exists and a new entry appears in the picker with no further wiring, which is what the header already promised for the menu.

## Where it lives

The picker is the ImGui `StartupView`, which is JUCE-free, so the new panel adds no JUCE. `tools/juce-gate.sh` is unchanged at 163 files / 8067 uses, and `MainComponent.cpp` stays at its recorded ceiling.

## Verification

Xvfb, fresh HOME: launch, NEW, the panel lists all four; picking **Band** goes to naming; saving writes a session whose tracks are `Kick / Snare / Drums OH / Bass / Gtr 1 / Gtr 2`. Screenshots of the panel and the result are in the session scratchpad.

Build clean with no new warnings, ctest 983/983.

## Test

`tests/session_templates.cpp` pins what the picker now depends on: every enum entry has a distinct non-empty label, applying any template leaves no track unnamed, and Band actually differs from Blank. README's test-case count moves to 1004 across 189 files; #538, #540, #545 and #553 touch the same line, so whichever lands last re-bumps.

## Manual

The first-launch paragraph now says NEW offers the starting points, and notes the menu route stays for later.

## Noticed while testing

On this branch, which is based on main and so predates #553, the walk hit #549 exactly as reported: the save-changes prompt appeared on NEW after the autosave had fired, on a session nothing had touched. Independent confirmation of that bug and of the fix in #553.


marc-korte:

Issue #558, opened 2026-09-09
[Windows] Keyboard shortcuts do nothing in the main window


marc-korte:
Found walking the demo path on Windows (#536).

## Setup

Windows 11 VM, no GPU. Build from `origin/main` (the `windows-tests.yml` artifact added in #552), run from `%LOCALAPPDATA%\dusk-walk` beside the Mesa `opengl32.dll` + `libgallium_wgl.dll` the MSI ships. App running, its window foreground, mixer drawn and responsive-looking.

## Steps

Send each of these to the foreground app window:

1. `Space` (Play / Stop)
2. `Ctrl+Q` (Quit)
3. `Ctrl+2` (Mixing stage)
4. `T` (show / hide timeline)
5. Assorted letters that are bound: `R`, `A`, `S`, `M`, `B`, `P`, `L`, `C`

## Expected

Per the keyboard reference in MANUAL.md, Space rolls the transport, Ctrl+2 switches to MIXING, T toggles the tape strip, Ctrl+Q quits.

## Actual

Nothing. The transport stays at `1.1.000` with Stop lit, the stage stays on RECORDING, the timeline does not toggle, the app does not quit, no ARM or SOLO lights up. Not one of the shortcuts does anything.

## Not a harness artefact

Key injection reaches the guest: with the same mechanism, the Windows key opens the Start menu, and typing into PowerShell works normally. Only the Dusk Studio window ignores keys.

The same source build under Xvfb on Linux responds to `Space` immediately: play lights green and the transport reads `2.2.467` three seconds later. So this is not a build or a shortcut-table problem.

## Caveat, and why it is filed anyway

This machine has no GPU. It runs on the Mesa software stack the MSI bundles, and this launch was on the fallback path where the startup dialog had already been marked unavailable for this display, so the main window is drawn by the plain path. I cannot separate "Windows" from "this display stack" without a second Windows machine, and someone should reproduce on real hardware before the fix is designed.

It is filed because the reachable behaviour is severe: with a mouse a user can still work, but every keyboard shortcut in the manual is dead, and on this machine no amount of clicking substitutes for `Space`.

## Screenshots

Transport before and after `Space` (identical, stopped at `1.1.000`); the Start menu opening under the same injection path, proving keys arrive; the Linux transport rolling after the same keystroke. In the session scratchpad.

marc-korte:
[...]

Issue #559, opened 2026-09-09
[Windows] First launch on a software-GL display hangs with a white window; the startup picker is then disabled for good


marc-korte:
Found walking the demo path on Windows (#536).

## Setup

Windows 11 VM, no GPU, so OpenGL comes from the Mesa build the MSI bundles. Build from `origin/main`, run beside `opengl32.dll` + `libgallium_wgl.dll` copied from the installed `C:\Program Files\Dusk Studio\bin`.

## What happens

**First launch, clean config.** The renderer resolves to `D3D12 (Microsoft Basic Render Driver) version=4.6 Mesa 26.2.0`. The app starts drawing the startup session picker and the window stays a blank white rectangle. It never paints and never recovers; the process has to be killed.

**Every launch after that.** The app comes up fine, but the header carries `Startup dialog unavailable on this display; opened the default session` and the session picker never appears again. `stderr` explains:

```
[Dusk Studio/startup] a previous run ended while the window was drawing its first
frame on GDI Generic; unavailable. Delete
C:\Users\marc\AppData\Roaming\Dusk Studio\startup-first-frame to try again
```

Deleting that latch and relaunching reproduces the white window, and the latch is written again. So the state is stable: launch one hangs, launch two onward is usable without a session picker.

## Reading

The latch is working as designed and is the reason the app is usable at all here. Two things still land on the user:

1. The very first launch hangs. Nothing on screen says so, and the remedy is Task Manager.
2. From then on there is no session picker, and the only way back is deleting a file named in a message a GUI user never sees. The recovery instruction is written to `stderr`, which on Windows goes nowhere unless the app was started from a console.

This is the first-run path, on the display stack the MSI itself ships, so it is not an exotic configuration: any Windows machine without a working GL driver (VM, RDP session, bare install before GPU drivers) lands here.

Related: #336 covered the notepad refusing a pre-OpenGL-3 display. This is the startup dialog on the same kind of display, and the failure is a hang rather than a refusal.

## Not filed as a blocker

The machine has no GPU and I could not try real Windows hardware. If it reproduces there, or over RDP, it should be raised: a first launch that has to be killed is the worst possible first impression.

## Screenshots

The white window on the hanging launch; the header on a subsequent launch showing the fallback message; the `stderr` text above. In the session scratchpad.

marc-korte:
Closing this: it was my harness, not the product.

[...]

PR #568, opened 2026-09-09
fix(ui): give the canvas keyboard focus when no startup picker opens

marc-korte:
Refs #558. It does **not** close it; see the correction at the end.

## Cause

`MainComponent::keyPressed` is where every shortcut in the manual lives, and JUCE only calls it when the canvas holds keyboard focus. Exactly one place granted that focus, on the startup picker's dismissal, and its comment says why:

```cpp
// main canvas so transport / edit shortcuts work without a stray click
// first (StartupDialog isn't an EmbeddedModal, so there's no
// focusRestoreTarget hand-back to lean on).
if (! safeThis->startupQuitRequested && safeThis->isShowing())
    safeThis->grabKeyboardFocus();
```

Every route that shows no picker skipped it: `DUSKSTUDIO_LOAD_SESSION`, `DUSKSTUDIO_SKIP_STARTUP_DIALOG`, `DUSKSTUDIO_CAPTURE_DIR`, and the one that matters to users, a display that cannot carry the panel. On that last one the app prints `Startup dialog unavailable on this display; opened the default session` and carries on with nothing focused, so JUCE hands key events to the peer's component, which is the window rather than the canvas, and they never route back down.

That is the shape of the Windows report in #558, and it is a real gap on its own account: any of those routes leaves a user with no working shortcut until they click the canvas.

## Evidence

I instrumented `keyPressed` on a scratch build and ran both routes under Xvfb.

Picker shown and dismissed: `b`, `t`, `ctrl+s`, `space` and `ctrl+q` all arrive, `ctrl+s` opens Save As, `space` rolls the transport, `t` toggles the timeline.

`DUSKSTUDIO_SKIP_STARTUP_DIALOG=1`, the same no-picker path Windows takes: **zero** key events reach `keyPressed`.

With this change on that path, the latched grab now runs and succeeds:

```
[FP] pending=1 showing=0     <- startup routing, too early to focus
[FP] pending=1 showing=1     <- parentHierarchyChanged, window on screen
[FP] grabbed self=1
```

Before it, `showing` was never true again at any point the code looked, because `resized()` does not fire a second time once the window is up.

## The fix

Latch the request in `focusMainCanvas()` and take it in `parentHierarchyChanged()`, which is what JUCE walks down every descendant when the window becomes visible. The startup routing in the constructor also collapses into one queued lambda with three outcomes rather than two branches and a silent third, which is what let the gap hide.

No new JUCE: the gate goes from 8067 to 8066 uses, and `MainComponent.cpp` sits one under its recorded ceiling.

## This is not the Windows fix

Retested on the installed MSI build on the VM, which draws the picker normally. With the picker shown and dismissed, `Space` does nothing and `T` leaves the screen pixel-identical, while the Windows key opens the Start menu on the same injection path. So Windows ignores keys even when the canvas has been given focus, and #558 stays open with that evidence recorded. What this PR fixes is the no-picker routes, which are reachable on every platform.

## What is not verified, and why

The user-visible half is not confirmed end to end. This box has no window manager, so on a WM-less Xvfb the JUCE window never receives X input focus on the no-picker route, and key events do not reach the process at all even with `xdotool windowfocus`. The picker route works there only because the panel is a separate native window that takes X focus and hands it back. So Linux can show that the focus grant now happens where it never did, and cannot show the keystroke arriving.

That check belongs on the VM, and needs a Windows build of this branch, which needs #552 (the artifact upload) merged first. I will run it and report on this PR as soon as that lands.

## Separately

`Ctrl+Q` reaches `keyPressed` and dispatches Quit correctly, and the app still does not exit. It behaves the same from **File > Quit**, so it is a quit defect rather than a shortcut one and is not touched here. I will file it separately rather than fold it in.

### DIFF
--- a/src/ui/MainComponent.cpp
+++ b/src/ui/MainComponent.cpp
@@ -1011,28 +1011,33 @@ MainComponent::MainComponent()
     // benchmarking the load path (the [Dusk Studio/Load] timing line ends
     // up in the parent terminal) and for scripted reproductions of
     // user-reported regressions.
-    if (const char* loadPath = std::getenv ("DUSKSTUDIO_LOAD_SESSION");
-        loadPath != nullptr && *loadPath)
+    // startupDialogPending was already set before setSize() above (so the
+    // scan-kicking resized() saw the gate); the routing here just decides what
+    // the first tick does. Capture mode suppresses the picker too - its modal
+    // would overlay the snapshots (mirrors the CAPTURE_DIR guard in the scan
+    // path).
+    //
+    // The last branch is not a no-op: only the picker's dismissal hands
+    // keyboard focus to the canvas, so a route that shows no picker leaves the
+    // window with nothing focused, and every shortcut in the manual is dead
+    // until the user happens to click the canvas.
     {
+        const char* loadPath = std::getenv ("DUSKSTUDIO_LOAD_SESSION");
+        const bool loadsSession = loadPath != nullptr && *loadPath;
+        juce::String pathStr (loadsSession ? loadPath : "");
+        const bool wantsPicker = ! loadsSession
+                              && std::getenv ("DUSKSTUDIO_SKIP_STARTUP_DIALOG") == nullptr
+                              && std::getenv ("DUSKSTUDIO_CAPTURE_DIR") == nullptr;
+
         juce::Component::SafePointer<MainComponent> safeThis (this);
-        juce::String pathStr (loadPath);
-        dusk::callAsync ([safeThis, pathStr]
+        dusk::callAsync ([safeThis, pathStr, loadsSession, wantsPicker]
         {
-            if (safeThis != nullptr)
-                safeThis->loadSessionFromJson (juce::File (pathStr));
-        });
-    }
-    else if (std::getenv ("DUSKSTUDIO_SKIP_STARTUP_DIALOG") == nullptr
-             && std::getenv ("DUSKSTUDIO_CAPTURE_DIR") == nullptr)
-    {
-        // startupDialogPending was already set before setSize() above (so the
-        // scan-kicking resized() saw the gate); just queue the dialog here.
-        // Capture mode suppresses the picker too - its modal would overlay
-        // the snapshots (mirrors the CAPTURE_DIR guard in the scan path).
-        juce::Component::SafePointer<MainComponent> safeThis (this);
-        dusk::callAsync ([safeThis]
-        {
-            if (safeThis != nullptr) safeThis->launchStartupDialog();
+            auto* const self = safeThis.getComponent();
+            if (self == nullptr) return;
+
+            if (loadsSession)      self->loadSessionFromJson (juce::File (pathStr));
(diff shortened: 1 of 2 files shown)


marc-korte:
Chased the quit observation from the PR body before filing anything, and it does not hold up.

Clean run on this branch: fresh HOME, picker path, `Escape` to skip, then `Ctrl+Q`. The window is gone and the process is gone. Quit works.

So the earlier "Ctrl+Q dispatches but the app stays up" was a confound of the runs it came from, both of which had a modal in play: one had just cancelled a Save As chooser, the other had been through New session and back. That matches the Linux leg of #536, where File > Quit exited cleanly on the packaged tarball.

Nothing to file. The quit paragraph in the PR description above is wrong and this comment supersedes it.

marc-korte:
Correction, and it matters for whether this can be merged as-is.

I retested on the installed MSI build on the VM (the one that draws the startup picker normally). With the picker shown and dismissed, so with the canvas focus grant having run, `Space` still does nothing and `T` leaves the screen pixel-identical. The Windows key opens the Start menu on the same injection path a second later, so keys reach the guest.

So this PR is not the fix for #558. It closes a real gap, the no-picker routes never granting the canvas focus, which is reachable on every platform and is worth having on its own. But the Windows report has another cause, and I have edited the description so this no longer claims `Closes #558`. The issue stays open with the new evidence on it.

I also closed #559 as my own harness artefact: the `windows-tests.yml` artifact is built without `DUSKSTUDIO_USE_WINDOWS_SOFTWARE_OPENGL`, so it never forces `GALLIUM_DRIVER=llvmpipe`, and with the MSI Mesa DLLs copied beside it Mesa picked the D3D12 backend the repo already records as unsafe. The shipped build is fine.

marc-korte:
VM confirmation, on the merged change.

Instrumented build of `main` (which now carries this fix), run on the Windows 11 VM with `GALLIUM_DRIVER=llvmpipe` and the app activated with `SetForegroundWindow` before any key was injected:

```
pid=6936 responding=True
foreground: 'Dusk Studio'

[DIAG/focus] pending=1 showing=0     <- startup routing, too early to focus
[DIAG/focus] pending=1 showing=1     <- parentHierarchyChanged, window on screen
[DIAG/focus] grabbed hasFocus=1      <- canvas holds focus
[DIAG/key]   code=32 cmd=0 shift=0   <- Space
[DIAG/key]   code=84 cmd=0 shift=0   <- T
[DIAG/key]   code=66 cmd=0 shift=0   <- B
```

Run on the no-picker route (`DUSKSTUDIO_SKIP_STARTUP_DIALOG=1`), which is the one this fix exists for. The transport reads `4.2.302` with play lit in the screenshot straight after, so the keys did not merely arrive, they did their job.

The Linux side of the story is unchanged: the grab happens where it never did before. Windows now shows the whole chain end to end.

Related: #558 is closed. Its evidence was mine and it was wrong; the foreground window during those tests was PowerShell, not the app. The gap this PR closes was real and is separate from that.

PR #569, opened 2026-09-09
feat(ui): open the shipped quickstart from the Settings menu

### LINKED ISSUE #533: [Docs] Quickstart page for the first five minutes, linked from the app and every package

marc-korte:
Closes #533
Refs #561, which adds the document this points at.

## What it does

**Settings > Quickstart** hands the shipped quickstart to the desktop's default handler. That is the same menu as Keyboard Shortcuts, Supporters and About, so no new menu surface.

## Finding the file

Resolved from the executable, not the working directory, because the working directory is wherever the launcher happened to be. The candidates, in order:

| Path | Covers |
|---|---|
| `<exe>/../QUICKSTART.txt` | Linux tarball (`DuskStudio/DuskStudio` under the root) and the Windows MSI (`bin/DuskStudio.exe`) |
| `<exe>/QUICKSTART.txt` | a flat layout |
| `<exe>/../Resources/QUICKSTART.txt` | inside a macOS bundle, if packaging later puts it there |
| `<exe>/../../../QUICKSTART.md` | the repo copy, so a build run out of the tree finds it |

These match `packaging/contents.txt` on #561's branch, which places `QUICKSTART.txt` at the install root on all three platforms.

macOS is worth calling out: the DMG carries the file next to `DuskStudio.app`, and dragging the app to Applications leaves it behind. Until packaging puts a copy in the bundle, a macOS user who installs the normal way gets the disabled entry. The Resources candidate is there so that becomes a packaging change rather than a code change.

## Missing file

The entry is disabled and reads `Quickstart  (not installed)`. No modal. A `PopupMenu` item has nowhere to hang a tooltip, so the label carries the reason.

## Platform seam

`executableDirectory()` and `openPathInDefaultApp()` join `duskstudio::platform` alongside the existing windowing primitives, one implementation per OS: `/proc/self/exe` and `xdg-open` (double-forked so the viewer reparents to init and leaves no zombie), `GetModuleFileNameW` and `ShellExecuteW`, `_NSGetExecutablePath` and `NSWorkspace`. No JUCE in any of them, and the gate is unchanged at 163 files / 8067 uses.

## Verification

Xvfb, all three states, screenshots in the session scratchpad:

- Document present: the entry is enabled and reads `Quickstart`.
- Document absent: greyed out, reads `Quickstart  (not installed)`.
- Clicking it with a stub `xdg-open` first on `PATH`: the stub records `/home/marc/projects/DuskStudio-worker-a/QUICKSTART.md`, so the resolution and the handoff both work end to end.

Build clean, no new warnings, ctest 980/980.

Windows and macOS are compile-only here. The Windows leg is the MSVC job; I have no way to click the menu on the VM until #558 is understood, and macOS is not mine to touch this round.

## Summary by CodeRabbit

- **New Features**
  - Added **Settings → Quickstart** to open the bundled one-page quickstart guide in the system’s default text viewer.
  - The menu entry locates the guide across supported installations and platforms.
  - When the guide is unavailable, the entry is disabled with an explanatory label.
  - Displays an alert if the guide cannot be opened by the system’s default application.

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -530,6 +530,8 @@ Only one stage is visible at a time, but the same engine drives all four. Switch
 - **MASTERING** swaps the console view for the mastering chain, including a file picker for loading a finished mix.
 - **AUX** swaps the console view for the four aux return lanes, with a full-width view of each lane's plugin chain.
 
+**Settings → Quickstart** opens the one-page quickstart that ships with the app, in whatever your system uses for text; it is greyed out and says so when that file is not installed.
+
 Press **Cmd/Ctrl+1 / 2 / 3 / 4** to jump straight to RECORDING / MIXING / MASTERING / AUX. The bank selector has the plain number keys **1 through 8**, so a modified digit changes stage and a plain digit changes the visible page of strips. Hovering any tab or bank button shows its shortcut, and **?** opens a full keyboard-shortcut list (also under **Settings → Keyboard Shortcuts**).
 
 Switching into or out of MASTERING force-stops the transport. The mix engine and the mastering engine cannot run at the same time. Changing stage also closes the notepad, saving it — the two cannot share the window.
--- a/src/ui/MainComponent.cpp
+++ b/src/ui/MainComponent.cpp
@@ -5346,8 +5346,17 @@ void MainComponent::menuItemSelected (int menuItemID, int /*topLevelMenuIndex*/)
         case kMenuSettingsQuickstart:
         {
             const auto quickstart = locateQuickstartDocument();
-            if (! quickstart.empty())
-                duskstudio::platform::openPathInDefaultApp (quickstart);
+            if (quickstart.empty()) break;
+
+            if (! duskstudio::platform::openPathInDefaultApp (quickstart))
+            {
+                // Nothing was launched, so say where the file is rather than
+                // leaving a menu click that silently did nothing.
+                const auto message = "Dusk Studio could not hand this file to a "
+                                     "default application: " + quickstart.string()
+                                   + ". Open it from there by hand.";
+                showDuskAlert (*this, "Could not open the quickstart", message.c_str());
+            }
             break;
         }
         case kMenuSettingsAbout:
--- a/src/ui/MainComponent.cpp
+++ b/src/ui/MainComponent.cpp
@@ -60,6 +60,7 @@
 #include "../foundation/PlanarBuffer.h"
 #include "../foundation/Text.h"
 #include <algorithm>
+#include <filesystem>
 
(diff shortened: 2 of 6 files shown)


marc-korte:
Corrected a real bug in this PR, found while building the DMG for the macOS leg of #536.

I had the filename wrong. `packaging/contents.txt` on main requires **`QUICKSTART.md`** on all three platforms, and `scripts/package-macos.sh` and `package-tarball.sh` install it under that name. This branch only looked for `QUICKSTART.txt` outside the repo tree, so the entry would have read `Quickstart  (not installed)` on every real install while passing every test I ran, because my dev-tree check happened to use the `.md` candidate.

The freshly built DMG, mounted:

```
DuskStudio.app
LICENSE
LICENSES.txt
QUICKSTART.md
```

`scripts/verify-package-contents.sh macos` on it: `macos package contents OK (5 paths checked)`.

Fixed by searching both names in each candidate root, `.md` first. The comment now points at the contract rather than at my assumption, and the `.txt` fallback is there so a rename in packaging degrades to still working rather than to a silently disabled menu item.

Re-verified under Xvfb with a `.md` present: the entry is enabled and reads `Quickstart`. Build clean, gate unchanged at 165 files / 8135 uses. Force-pushed.

The macOS observation from the PR description stands and is now confirmed on a real DMG: `QUICKSTART.md` sits next to `DuskStudio.app` at the DMG root, so dragging the app to Applications leaves it behind and the entry will read `(not installed)` for a normal macOS install until packaging puts a copy inside the bundle. The `Resources` candidate is already in the list, so that stays a packaging change.

PR #572, opened 2026-09-09
fix(single-instance): ride out a full handoff backlog instead of running unattached

### LINKED ISSUE #567: [Single instance] A full handoff backlog makes a launch run unattached and drop its command line
Seen while validating #566. Filed so it is tracked rather than rediscovered.

## The failure

```
tests/single_instance_socket_lifecycle.cpp:318: FAILED:
  REQUIRE( std::count (primary.begin(), primary.end(), 1) == 1 )
with expansion:
  3 == 1
```

Three of eight simultaneous launches each returned true from `acquire()`, where the test requires exactly one.

## Reproduction

```bash
cmake --build build-tests --target dusk-studio-tests -j3
for i in $(seq 1 42); do
    ./build-tests/tests/dusk-studio-tests \
        "simultaneous single-instance launches elect exactly one owner" || echo "FAILED run $i"
done
```

**One failure in 42 runs.** It failed on run 4, then passed 37 consecutive runs afterwards. Rare, not systematic.

## Where it reproduced

On `origin/main` plus only the `Deliveries` mutex from #566. **#541 was not in the tree**, so this is not caused by the POSIX handoff acknowledgement that PR adds. Recording that explicitly because #541 changes this exact path and would otherwise be the obvious suspect the next time this fires.

It is also distinct from the TSan data race in #565, which was in the test harness's vector and is fixed by #566. This one is about the election result, not the delivery bookkeeping.

## Two candidate readings, undecided

1. **The assertion is stronger than the contract.** `acquire()` returns true both for "you are the primary" and for "run unattached", the latter when a launch cannot reach or cannot confirm handoff to a live primary. Several launches legitimately returning true under contention would then be correct behaviour and the test would be asserting something `acquire()` never promised. The fix would be to assert on ownership rather than on the return value.

2. **A real double primary under load.** Eight launches contending should elect one owner and hand off seven times. Three running means the election or the handoff genuinely failed twice under contention. That is two instances writing one session directory, each with its own engine and autosave loop, which is data loss on a shared session.

Reading 2 is why this is labelled `1.0`: it has to be resolved before release even if the resolution turns out to be "the test asserts the wrong thing".

## Not investigated further

Deliberately stopped here rather than guessing. Distinguishing the two needs instrumenting which branch of `acquire()` each thread took, which is a piece of work in its own right.

marc-korte:
Closes #567

#567 was filed as a flake: three of eight simultaneous launches returned `true` from `acquire()` where the test requires exactly one. The question was whether the test asserted more than the contract, or whether there was a real double primary.

Neither, exactly. There is one primary every time, and there is a real bug.

## What the instrumentation showed

Every `acquire()` return path on the POSIX arm was tagged and the election test looped. **12 failures in 6205 runs.** In all twelve, exactly one launch bound the socket and started a listener. The `SlotLock`/`flock` election is sound: every extra `true` was a `reportUnattached` return from a thread that had taken the lock, probed the slot as `Live`, correctly declined to bind, and then failed to hand off.

The failing threads all had the identical trace:

```
handover connect errno=11           <- EAGAIN
handover poll revents=20            <- POLLOUT|POLLHUP
handover writeAll failed errno=107  <- ENOTCONN
handover=Failed errno=107
RESULT=UNATTACHED errno=107
```

## Root cause

1. `startListener` calls `::listen (fd, 8)`. Eight racing launches open up to fourteen sockets against it (a `probeSlot` connect plus a `handOver` connect each), so the backlog transiently overflows.
2. On a full Unix-domain backlog, `connect()` returns **EAGAIN**. That is not "in progress" - nothing was started.
3. `handOver` lumped `EAGAIN`/`EWOULDBLOCK` in with `EINPROGRESS` and polled for `POLLOUT`. An unconnected Unix socket answers `POLLOUT|POLLHUP` immediately and `getsockopt(SO_ERROR)` returns 0, because no error was ever recorded, so the code concluded it was connected.
4. The write on that never-connected socket failed **ENOTCONN**, giving `Handoff::Failed`.
5. `Failed` is not `Refused`, so `acquire()` left the retry loop, reported unattached, and returned `true` - **dropping the command line it was carrying**. On the desktop that is a session opened while the app is starting landing in a second window that does not have the file.

`probeSlot` misread the same errno. It happened to land on `Slot::Live`, which is the safe answer, but by accident.

## The change

- Poll for connect completion only on `EINPROGRESS`/`EALREADY`/`EINTR`.
- Retry a full backlog instead. It gets the whole `kDeliveryBudgetMs` rather than the short `kRefusalRetryMs`, because unlike a refusal it proves someone is listening; a dead primary still must not add the full delivery timeout to startup.
- A full backlog is never reported as `Refused`. `Refused` is what licenses unlinking the slot, and unlinking a live primary's socket is how you get two of them.
- `probeSlot` returns `Slot::Live` on EAGAIN deliberately, with the reason recorded.

The `listen` backlog stays at 8. Raising it would only narrow the window and would weaken the regression test.

## Regression test

`a handoff waits out a full primary backlog` binds a socket at the slot path, fills its backlog until a connect is actually turned away (the kernel is free to round the backlog up, so it fills until EAGAIN rather than trusting the number passed to `listen`), and drains it 150 ms later. The handoff must survive the saturation and arrive.

- **Before the fix**: `REQUIRE_FALSE (ranUnattached)` fails with `!true` - the launch gave up and ran unattached.
- **After the fix**: passes, and the payload arrives intact.

The drainer writes a one-byte acknowledgement before closing, which `main` ignores but the POSIX ack path in #570 will read, so the test does not become a silent-listener case when that lands.

## Verification

| | before | after |
|---|---|---|
| `simultaneous single-instance launches elect exactly one owner`, 2500 iterations | **15 failures** | **0 failures** |

- `ctest --test-dir build-tests`: 983/983 pass.
- `tools/juce-gate.sh`: clean, 163 coupled files / 8066 uses, no allowlist change.
- `scripts/run-selftest-xvfb.sh`: 15 passed, 0 failed. (The UMC1820 direct-open probe reports the device held by PipeWire, as it does on `main`.)
- README test-case count updated 1002 -> 1003.

Relabelled #567 `1.0-blocker` and retitled it to describe the handoff bug; the flake history stays in the issue body.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1032 Catch2 test cases across 193 test source files. Linux
+The C++ suite declares 1033 Catch2 test cases across 193 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
@@ -124,7 +124,7 @@ src/
   session/     # Session model + JSON serialisation
   ui/          # MainComponent, ConsoleView, channel/aux/master strips, mastering view
   util/        # Native log storage + CrashHandler signal reports
-tests/         # 1032 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
+tests/         # 1033 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
 packaging/     # .desktop, AppStream, MIME, macOS bundle — for tarball + DMG builds
 DuskStudio.md  # authoritative product spec
 MANUAL.md      # end-user manual (Pandoc-buildable to PDF via docs/build-pdf.sh)
--- a/src/util/SingleInstance.cpp
+++ b/src/util/SingleInstance.cpp
@@ -529,8 +529,7 @@ Slot probeSlot (const sockaddr_un& addr, socklen_t len)
     if (::connect (fd, (const sockaddr*) &addr, len) != 0)
     {
         result = errno;
-        if (result == EINPROGRESS || result == EALREADY || result == EINTR
-            || result == EAGAIN || result == EWOULDBLOCK)
+        if (result == EINPROGRESS || result == EALREADY || result == EINTR)
         {
             const auto deadline = std::chrono::steady_clock::now()
                                 + std::chrono::milliseconds (kProbeBudgetMs);
@@ -547,6 +546,10 @@ Slot probeSlot (const sockaddr_un& addr, socklen_t len)
     ::close (fd);
 
     if (result == 0) return Slot::Live;
+    // A Unix-domain connect answers EAGAIN when the backlog is full, which only
+    // a listening socket can be, so this is the one refusal-shaped error that
+    // proves the slot is occupied rather than abandoned.
+    if (result == EAGAIN || result == EWOULDBLOCK) return Slot::Live;
     if (result == ECONNREFUSED) return Slot::Orphaned;
     if (result == ENOENT) return Slot::Vacant;
     return Slot::Unknown;
@@ -611,9 +614,14 @@ Handoff handOver (const sockaddr_un& addr, socklen_t len, const std::string& pay
         else
         {
             connectErr = errno;
+            // Only EINPROGRESS and its relatives mean a connect is under way.
+            // EAGAIN from a Unix-domain connect means the backlog is full and
+            // nothing was started: poll() then answers POLLOUT|POLLHUP for a
+            // socket that never connected, SO_ERROR is clear because no error
+            // was ever recorded, and the write that follows fails ENOTCONN -
+            // costing the launch the command line it was carrying.
             if (connectErr == EINPROGRESS || connectErr == EALREADY
-                || connectErr == EINTR || connectErr == EAGAIN
-                || connectErr == EWOULDBLOCK)
+                || connectErr == EINTR)
             {
                 const int revents = pollUntil (fd, POLLOUT, deadline);
                 if ((revents & POLLNVAL) == 0 && revents != 0)
@@ -658,7 +666,8 @@ Handoff handOver (const sockaddr_un& addr, socklen_t len, const std::string& pay
                                           : Handoff::Submitted;
         }
         ::close (fd);
-        if (connectErr != ECONNREFUSED)
+        const bool backlogFull = connectErr == EAGAIN || connectErr == EWOULDBLOCK;
+        if (connectErr != ECONNREFUSED && ! backlogFull)
         {
             failureErrno = connectErr;
             return Handoff::Failed;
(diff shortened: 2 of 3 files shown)


marc-korte:
Pre-checked the interaction with #570 (`batch/1.0-fixes`), since the #541 acknowledgement lands in this same function.

`src/util/SingleInstance.cpp` cherry-picks onto that branch **with no conflict**, and the merged result is correct on both counts:

- `connected` is now true only for a socket that genuinely connected, so the `writeAll` failure path stops being reachable via a full backlog.
- A silent listener still returns **`Submitted`, not `Failed`**: `pollUntil (fd, POLLIN, deadline)` times out, `failureErrno` is set to `ETIMEDOUT`, `acked` stays false, and the tail returns `Handoff::Submitted`. My change does not touch that block.

The only conflicts are additive and trivial - `README.md` (both bump the test-case count) and `tests/single_instance_socket_lifecycle.cpp` (both add a TEST_CASE in the same region). I will resolve them when whichever of the two lands second is rebased.

marc-korte:
Rebased onto `main` at 7d168644, after #570 and #571 merged.

- `src/util/SingleInstance.cpp` rebased with **no conflict**, exactly as the dry run predicted.
- `tests/single_instance_socket_lifecycle.cpp`: both branches added a TEST_CASE in the same place. Kept both - #570's `a handoff nobody acknowledges leaves the launch unattached` and this PR's `a handoff waits out a full primary backlog`.
- `README.md`: recounted from the tree rather than reusing either side's number. 1032 -> **1033** test cases across 193 source files.

Re-confirmed on the merged file that a **silent listener still returns `Submitted`, not `Failed`**: `pollUntil (fd, POLLIN, deadline)` times out, `failureErrno` becomes `ETIMEDOUT`, `acked` stays false, and the tail returns `Handoff::Submitted`. The two changes are disjoint - #570's acknowledgement is inside `if (connected)`, and this fix is about when `connected` may be believed in the first place.

Post-rebase locally: `ctest` 1013/1013, `tools/juce-gate.sh` clean with no allowlist change, all 15 `[single-instance]` cases pass.

marc-korte:
macOS CI was red on my own portability bug, now fixed.

```
tests/single_instance_socket_lifecycle.cpp:463:57: error: use of undeclared identifier 'SOCK_NONBLOCK'
```

`SOCK_NONBLOCK` is a Linux flag; the fill loop now does `::socket` + `fcntl (F_SETFL, O_NONBLOCK)`, which is what the production `openSocket` does for the same reason.

That prompted the obvious follow-up question, and the answer changes the test's scope. I ran a standalone probe on both platforms - bind, `listen (fd, 1)`, then nonblocking connects until one is turned away:

| | result |
|---|---|
| Linux | queues 2, then **EAGAIN (11)** |
| macOS arm64 | queues 1, then **ECONNREFUSED (61)** |

So a full Unix-domain backlog is simply not the same condition on macOS, and the state this test builds cannot exist there. The TEST_CASE and its two helpers are now behind `#if defined (__linux__)`, matching the existing gate on `captureStderr` in the same file, with the reason recorded in a comment. The production change stays POSIX-wide and is inert on macOS, where EAGAIN never arises from this cause.

Local after the fix: `ctest` 1013/1013, `tools/juce-gate.sh` clean, all 15 `[single-instance]` cases pass. The production file is byte-identical to the version that ran 2500 election iterations with zero failures, so that measurement stands.

Separately, the macOS probe result points at a defect on that platform that is **not** in this PR's scope - I have written it up for the maintainer rather than widening this change.

PR #574, opened 2026-09-09
fix(single-instance): mark the slot with an owner lock instead of trusting a refusal

### LINKED ISSUE #573: [Single instance] On macOS a full backlog reads as an orphaned slot and elects a second primary
Found while fixing #567. That bug was Linux-only; this is the same family on macOS, and it is worse, because what is lost is the session rather than the command line.

## The platform difference

A standalone probe: bind an `AF_UNIX` socket, `listen (fd, 1)`, then nonblocking `connect()` until one is turned away.

| platform | queued before refusal | errno when the backlog is full |
|---|---|---|
| Linux (6.12, amd64) | 2 | **EAGAIN (11)** |
| macOS 14 arm64 | 1 | **ECONNREFUSED (61)** |

Linux distinguishes "backlog full" from "nobody is listening". macOS does not: both are `ECONNREFUSED`.

## Why that is a double primary

`probeSlot` in `src/util/SingleInstance.cpp` decides liveness purely from the connect result:

```cpp
if (result == ECONNREFUSED) return Slot::Orphaned;
```

`Orphaned` is the one state that licenses removing the slot entry. Follow `acquire()` on macOS with a live primary whose backlog is momentarily full, all of it under `SlotLock` so nothing else can intervene:

1. `probeSlot` connects, gets `ECONNREFUSED` because the backlog is full, and answers `Slot::Orphaned`.
2. `::unlink (path)` removes the **live primary's** socket.
3. `state != Slot::Live`, so the launch opens a socket and `bind`s - which now succeeds, because the entry it would have collided with has just been deleted.
4. `startListener` succeeds and `acquire()` returns `true`.

The result is two primaries. The original is still running and still listening, on a socket no launch can reach any more; the new one owns the path. Both have their own engine and their own autosave loop over one session directory.

The comment on `SlotLock` already names this exact outcome as the thing the lock exists to prevent:

> two launches racing after a crash would then each drop what the other just bound, and both would run as primary over one session directory

The lock does prevent the race it describes. It cannot help here, because on macOS the answer `probeSlot` returns from inside the lock is wrong.

## Reachability

Narrower than the Linux flake, but real. BSD rounds `so_qlimit` up, so `listen (fd, 8)` has genuine headroom and the macOS election test has not flaked. It needs enough simultaneous launches to fill that backlog - the same shape as the eight-way race in `simultaneous single-instance launches elect exactly one owner`, which is exactly what opening a folder of sessions from Finder does.

The macOS leg of #536 has not been walked yet, so this has had no manual exposure either.

## Fix direction

`probeSlot` cannot decide liveness from a connect result on macOS, so liveness has to be made explicit and the same on both platforms: the primary takes a long-lived advisory lock on a companion file (`<slot>.owner`) after a successful `bind`, holds it for its whole lifetime, and releases it only in teardown - or has the kernel release it on a crash. This is separate from the existing `SlotLock`, which is a short critical section around mutations of the slot entry.

`probeSlot` then answers `Live` when a nonblocking `LOCK_EX` on the owner file fails with `EWOULDBLOCK`, and `Orphaned` only when it can take that lock (releasing it again) *and* the connect is refused. Unlinking stays licensed by `Orphaned` alone. The owner file itself is never unlinked, for the same reason the existing lock file never is.

Windows has its own contract and is out of scope.

marc-korte:
Closes #573

Found while fixing #567. That bug was Linux-only; this is the same family on macOS, and worse, because what is lost is the session rather than the command line.

**Stacked on #572** - it touches the same function and I did not want to hand-resolve a conflict in this particular code. The first commit here is #572's; review only `14a0564c`. I will rebase to a single commit once #572 merges.

## The platform difference

A standalone probe on both hosts: bind, `listen (fd, 1)`, then nonblocking `connect()` until one is turned away.

| platform | queued before refusal | errno when the backlog is full |
|---|---|---|
| Linux 6.12 amd64 | 2 | **EAGAIN (11)** |
| macOS 14 arm64 | 1 | **ECONNREFUSED (61)** |

Linux distinguishes "backlog full" from "nobody is listening". macOS does not.

## Why that elected a second primary

`probeSlot` decided liveness purely from the connect result, and `ECONNREFUSED` mapped to `Slot::Orphaned` - the one state that licenses removing the slot entry. On macOS, with a live primary whose backlog was momentarily full, all of this ran under `SlotLock` so nothing could intervene:

1. `probeSlot` gets `ECONNREFUSED` from the full backlog and answers `Orphaned`.
2. `::unlink (path)` removes the **live primary's** socket.
3. `state != Slot::Live`, so the launch binds - which now succeeds, because the entry it would have collided with was just deleted.
4. `startListener` succeeds, `acquire()` returns `true`.

Two primaries, each with its own engine and autosave loop over one session directory. The original is still listening, on a path no launch can reach.

`SlotLock`'s own comment names this outcome as the thing it exists to prevent. It does prevent the race it describes; it could not help here, because the answer `probeSlot` returned from inside the lock was wrong.

## The change

Liveness is now observable rather than inferred. The primary takes an advisory lock on `<slot>.owner` after it binds and holds it for its whole life, so the kernel drops it on a crash and no stale file can outlive the process the way the socket entry does.

- `startListener` takes the lock after `listen` and stores the descriptor on the `Listener`. A launch that cannot take it does **not** become primary - the mark is the only thing stopping the next launch from clearing this one.
- `probeSlot` answers `Orphaned` only when the connect is refused **and** the owner lock can be taken. Everything else that is refusal-shaped reads as `Live`. It fails closed: a lock that cannot be opened reads as owned.
- `teardown` drops the lock last, after the entry this process bound is gone, so no launch sees the slot unowned while a dead socket is still at the path. The socket unlink moved into `unlinkOwnSocket` so the lock is released on every teardown path, including the ones that return early.
- The owner file is never unlinked, for the reason the slot lock file is not (same residue-1 reasoning).

This is deliberately not the `SlotLock`: that is a short critical section every launch takes and releases around mutations of the slot entry, and a launch waiting on it would block forever if it doubled as the liveness mark.

Windows has its own contract and is untouched.

## Regression test

`a refusal from an owned slot leaves the socket alone` binds a socket that never listens (so every connect is refused) while holding the owner lock, then calls `acquire()` and asserts the entry is still the same inode. No Linux-only mechanics, so it runs on both platforms.

Proven on **macOS arm64** with a standalone harness linking only `SingleInstance.cpp`, run against this commit and its parent:

| | inode before | inode after | verdict |
|---|---|---|---|
| without the fix | 1785181 | 1785183 | **the live socket was replaced** |
| with the fix | 1785165 | 1785165 | unchanged, launch runs unattached |

On Linux the same test fails before the fix with `1094519 == 1094517` and passes after.

## Verification

- `ctest --test-dir build-tests`: 1014/1014.
- `tools/juce-gate.sh`: clean, no allowlist change.
- `scripts/run-selftest-xvfb.sh`: 15 passed, 0 failed.
- `cmake --build build -j3`: clean.
- README test-case count 1033 -> 1034.
- macOS full build + ctest running on the build node; CI's `Build + tests (clang Release, macOS arm64)` covers the same ground on the gate.

One aside for the record: two `DpAligner` cases failed once each under `ctest -j3` mid-run. Not related to this change - the box's 4 GB `/tmp` was at 100%, and those cases write ~23 MB WAVs. After clearing space, 1014/1014 twice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1033 Catch2 test cases across 193 test source files. Linux
+The C++ suite declares 1034 Catch2 test cases across 193 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
@@ -124,7 +124,7 @@ src/
   session/     # Session model + JSON serialisation
   ui/          # MainComponent, ConsoleView, channel/aux/master strips, mastering view
   util/        # Native log storage + CrashHandler signal reports
-tests/         # 1033 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
+tests/         # 1034 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
 packaging/     # .desktop, AppStream, MIME, macOS bundle — for tarball + DMG builds
 DuskStudio.md  # authoritative product spec
 MANUAL.md      # end-user manual (Pandoc-buildable to PDF via docs/build-pdf.sh)
--- a/src/util/SingleInstance.cpp
+++ b/src/util/SingleInstance.cpp
@@ -68,6 +68,7 @@ struct Listener
 {
     std::thread thread;
     std::atomic<int> listenFd { -1 };
+    std::atomic<int> ownerFd { -1 };
     std::atomic<bool> unlinked { false };
     int wakeFds[2] = { -1, -1 };
     std::string socketPath;
@@ -235,6 +236,47 @@ int acceptPeer (int fd)
 #endif
 }
 
+// Liveness has to be something a launch can observe directly, because a connect
+// result cannot carry it: macOS answers ECONNREFUSED both for a primary that
+// died and for one whose listen backlog is momentarily full, and acting on the
+// second as though it were the first deletes a live primary's socket and elects
+// a second one over the same session directory. The owner lock is held by the
+// primary for its whole life, so the kernel drops it on a crash and no stale
+// file can outlive the process the way the socket entry does.
+//
+// This is deliberately not the SlotLock: that one is a short critical section
+// every launch takes and releases around mutations of the slot entry, and a
+// launch waiting on it would block forever if it doubled as the liveness mark.
+// Like that lock's file, this one is never unlinked: removing it while a launch
+// holds it open leaves that launch locking an unlinked inode while another
+// locks a fresh one at the same name, and both then read the slot as theirs.
+std::string ownerLockPath (const std::string& socketPath)
+{
+    return socketPath + ".owner";
+}
+
+int takeOwnerLock (const std::string& socketPath)
+{
+    const int fd = ::open (ownerLockPath (socketPath).c_str(),
+                           O_RDWR | O_CREAT | O_CLOEXEC, 0600);
+    if (fd < 0) return -1;
+    if (::flock (fd, LOCK_EX | LOCK_NB) == 0) return fd;
+    ::close (fd);
+    return -1;
+}
+
+// Fails closed: a lock we cannot open or cannot reason about reads as owned, so
+// the caller leaves the slot entry alone rather than clearing one that may
+// still be serving.
+bool ownerLockIsFree (const std::string& socketPath)
+{
+    const int fd = takeOwnerLock (socketPath);
+    if (fd < 0) return false;
+    ::flock (fd, LOCK_UN);
+    ::close (fd);
+    return true;
+}
+
 // The invariant every launch relies on: the directory entry at the slot path is
 // created, made listenable, and removed only under this lock. A stat before the
 // unlink cannot stand in for it, because the entry can be replaced between the
@@ -286,14 +328,8 @@ private:
     int fd = -1;
 };
 
-// Closing the listening fd is what makes a later launch fall back to acting as
-// primary; leaving it bound but unserviced would strand every launch after it.
-// Both exchanges keep this safe to call from the listener thread and from
-// release() without double-closing.
-void teardown (Listener& l)
+void unlinkOwnSocket (Listener& l)
 {
-    const int fd = l.listenFd.exchange (-1);
-    if (fd >= 0) ::close (fd);
     if (l.unlinked.exchange (true) || ! l.socketIdentified) return;
 
     const SlotLock lock (l.socketPath);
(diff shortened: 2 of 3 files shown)


marc-korte:
## Confirmed: `Refused` alone cannot unlink or rebind

Traced from the code. `handOver` returning `Refused` does exactly two things in `acquire()`:

```cpp
        const Handoff outcome = handOver (addr, len, payload, handoffErrno);
        if (outcome == Handoff::Delivered) return false;
        // Anything short of a refusal may be a primary we could not talk to.
        // Run unattached rather than clear its slot: unlinking here promotes
        // this launch to a second primary, and two of those contend for one
        // audio device and one session directory.
        if (outcome != Handoff::Refused) break;
        handoffErrno = ECONNREFUSED;
```

It records an errno for the eventual message and lets the loop run again. No unlink, no bind on that path.

There are exactly two `unlink` calls reachable in `acquire()`, and both sit behind a **fresh** `probeSlot` verdict taken at the top of the next iteration, under `SlotLock`:

```cpp
                const Slot state = probeSlot (path, addr, len);
                if (state == Slot::Orphaned && ::unlink (path.c_str()) != 0 && errno != ENOENT)
```

```cpp
                if (state != Slot::Live)
                {
                    ...
                    if (::bind (fd, (const sockaddr*) &addr, len) == 0)
                    {
                        if (! startListener (fd, path, std::move (onCommandLine)))
                        {
                            ...
                            ::unlink (path.c_str());
```

The second is reachable only after a successful `bind`, which requires `state != Slot::Live` from that same verdict. And `Orphaned` is now issued only under the owner lock:

```cpp
    if (result != ECONNREFUSED) return Slot::Unknown;
    return ownerLockIsFree (path) ? Slot::Orphaned : Slot::Live;
```

So on a live macOS primary with a full backlog: `handOver` gives `Refused`, the loop re-probes, the owner lock is held, `probeSlot` answers `Live`, and both branches are skipped. The launch exits the loop and reports unattached. There is no hole to close.

**The existing regression test already drives this exact path.** `a refusal from an owned slot leaves the socket alone` binds a socket that never listens, so every connect is refused: attempt 0 probes `Live`, hands off, gets `Refused` after `kRefusalRetryMs`; attempt 1 re-probes `Live` and gets `Refused` again; the loop ends unattached. Asserting the inode is unchanged is the assertion that no unlink happened anywhere in that sequence.

## One thing the review question did turn up

The comment above that `probeSlot` call was left asserting the opposite of the new rule:

```cpp
                // A refusal seen here is final: a live primary completed
                // listen() before it released the lock, and nothing can create
                // or remove the entry while we hold it.
```

"A refusal seen here is final" is precisely the belief this PR removes. Replaced with what actually holds - the lock freezes the entry for the life of the verdict, and only the owner lock separates a dead slot from a refusing one. Comment-only, pushed as `d770c401`.

## macOS build node

| | result |
|---|---|
| `cmake --build build-tests --target dusk-studio-tests -j3` | clean, 0 errors |
| `ctest --test-dir build-tests -j3` | **987/987** |
| `[single-instance]` | 14/14, 157 assertions |
| election test, 600 iterations | **0 failures** |

Those ran on `14a0564c`; the only delta since is the comment above. Linux after the amend: `ctest` 1014/1014, `juce-gate.sh` clean.

Note for anyone reproducing on that box: `-DDUSK_PLUGINS_PATH=../plugins-main` fails configure with `Cannot find source file: ../plugins-main/plugins/multi-comp/multicomp.cpp`, because the relative path resolves against `tests/` rather than the repo root. An absolute path works.

marc-korte:
Flattened onto `main` at 405d5ebd now that #572 has merged. `edb4f9f3` is the single commit on the branch; git skipped #572's commit as already applied, so nothing needed resolving.

README recounted from the merged tree rather than carried over: 1034 across 193 source files, which is what the tree already said, so the line is unchanged.

Green locally on the flattened branch: `ctest` 1014/1014, `tools/juce-gate.sh` clean with no allowlist change, build clean.

Issue #578, opened 2026-09-10
[UI] The transport-bar notices are painted underneath the bank selector


marc-korte:
Found while verifying the backend-fallback notice added for #576. It affects the existing no-input notice from #554 the same way, and predates that work.

## What happens

`TransportBar::paint` draws the notice centred in the bar:

```cpp
auto notice = getLocalBounds().withSizeKeepingCentre (
    std::min (kNoticeW, getWidth()), std::min (20, getHeight() - 6));
```

`MainComponent` lays the bank selector out over the same strip of screen. `MainComponent.cpp` computes the band from the transport bar's own geometry and centres the bank cluster in it:

```cpp
const int bankBandLeft  = clockRight + 12;
const int bankBandRight = utilityLeft - 12;
...
const int groupX = std::clamp (stageCentreX - groupW / 2, bankBandLeft, ...);
```

The bank buttons are children of `MainComponent`, so they paint after the transport bar. The notice ends up underneath them.

## Evidence

At 3830x1180 with two bank buttons, the notice reads:

```
PipeWire un[ BANK 1 (1-18) ]A[ BANK 2 (19-24) ]gs > Audio.
```

At 1600x1000, where the bank row is eight buttons wide, the band is full and the notice is invisible apart from slivers of its background between buttons. Screenshots are in the session scratchpad.

## Why it has not been noticed

`docs/images/rec-02-no-input-notice.png` looks correct because the capture harness snapshots the transport bar component on its own:

```cpp
snapshotComponent (transportBar.get(), outDir, "rec-02-no-input-notice.png", 250);
```

A component snapshot does not include a sibling that overlaps it on screen, so the figure shows a notice nobody running the app in a banked session can read.

## Scope

Any session that needs banking, which is the default 24-strip session at any ordinary window size. Both notices are affected:

- "No input device. Choose one in Settings > Audio." (#554)
- "PipeWire unavailable - using ALSA. Change it in Settings > Audio." (#576)

Both exist to explain why the next thing the user tries will not work, so being unreadable defeats them.

## Fix shape

This wants an owner decision rather than a patch in `TransportBar::paint`, because the two components are laid out by different parents:

- give the notice its own row or a reserved slot that the bank cluster's band excludes, or
- have `MainComponent` draw the notice, since it already knows both rectangles, or
- keep it in `TransportBar` but let it claim the band and have `MainComponent` skip the bank cluster while a notice stands.

[...]

PR #579, opened 2026-09-10
fix(pipewire): wait for the graph's first cycle past a device waking up

### LINKED ISSUE #576: [Audio/Linux] First launch fails to open PipeWire with "no usable quantum (0)" and silently falls back to ALSA
Found re-walking the demo path on a packaged build (#536), step 1: does first launch pick a sane audio device.

## Setup

Packaged tarball built from `main` at 76571220 with `scripts/package-tarball.sh`, installed with `./install.sh` into a scratch `HOME`, run on a private Xvfb display. Host is a stock PipeWire desktop: `pipewire` 1.4.6 and `pipewire-pulse` running, no `~/.config/pipewire` overrides, `clock.quantum` 1024, `clock.max-quantum` 2048, rate 48000.

## What happens

On the first launch, with no `audio-device.xml` yet, the PipeWire open fails and the app falls back to ALSA:

```
[Dusk Studio/AudioEngine] device-manager init reported: PipeWire delivered no usable quantum (0)
[Dusk Studio/ALSA] opened "HDA Intel PCH, ALC700 Analog" rate=48000 period=1024 periods=3 bits=32 int out=2ch in=0ch (active out=2 in=0)
[Dusk Studio/ALSA] opened "HDA Intel PCH, ALC700 Analog" rate=48000 period=1024 periods=3 bits=32 int out=2ch in=2ch (active out=2 in=2)
[Dusk Studio/AudioEngine] first launch: selected input device "HDA Intel PCH, ALC700 Analog" alongside output "HDA Intel PCH, ALC700 Analog".
```

Reproduced four times out of four, deleting `~/.config/Dusk Studio/audio-device.xml` between runs.

Once the user touches anything in Audio Settings, `audio-device.xml` records `"deviceType": "ALSA"` and later launches go straight to ALSA without trying PipeWire at all, so the message never appears again and the choice sticks.

## Reading

The message comes from `PipeWireAudioIODevice::open`. The filter reaches `PW_FILTER_STATE_STREAMING` (otherwise the error would be the "did not start streaming" one immediately above it), then the code waits up to 200 ms for `negotiatedQuantum` to become non-zero and gives up with 0, meaning `onProcess` never ran in that window. The comment there says treating this as a failed open is deliberate, and it is the right call given the oversized-block guard; the problem is that it is happening at all on an ordinary desktop.

Effect on the user: PipeWire is the primary Linux backend, and on a stock PipeWire machine a first launch quietly ends up on ALSA `hw:0,0` instead, which is an exclusive open. Nothing in the UI says which backend was chosen or that a fallback happened; the only trace is stderr, which a user launching from the app menu never sees.

## Not a blocker

The demo path completed end to end over the ALSA fallback: record, playback, overdub and bounce all worked, and the Audio Settings panel lets the user pick the backend by hand afterwards. Filing at `1.0` rather than `1.0-blocker` on that basis.

## Not determined

Whether the quantum never arrives because the graph is idle and has not ticked, whether 200 ms is simply too short on this machine, or whether the node is linked to a driver that is not running. One machine, one PipeWire version. Worth a second Linux box before deciding on the fix.

marc-korte:
Closes #576

## Cause

`PipeWireAudioIODevice::open` waited 100 x 2 ms for `onProcess` to report the graph's real quantum and failed the open with `no usable quantum (0)` when no cycle arrived. `PW_FILTER_STATE_STREAMING` says our node was started, not that the graph has run. `pw-dump --monitor` during a failing open showed all four links active, the target sink running, and the elected driver still suspended. WirePlumber only resumes a suspended device node after the links go active, and the wait was racing that resume.

Measured with a temporary timestamp in the wait, first cycle after STREAMING on PipeWire 1.4.6:

| target device before launch | first cycle |
|---|---|
| already running (4 runs) | 1, 3, 6, 10 ms |
| resuming from suspend (4 runs) | 268, 270, 267, 270 ms |

Every device node is suspended on a launch with no saved device to reopen, which is why this only ever showed on a first launch. Once `audio-device.xml` recorded the ALSA fallback the message never appeared again.

## Change

- The wait is a 2 s deadline polled at 1 ms. The comment carries the measurement. The connect handshake above it already allows longer on the same thread, so this is not a new block.
- The quantum store/load pair is release/acquire.
- A one-line transport-bar notice when the app fell back off the preferred backend on a launch with no saved device state. `startupDeviceMessage` keeps the saved-intent case, so a user who deliberately saved ALSA never sees it. Pure helper with four Catch2 cases.

## Verification

Four cold launches on Xvfb with both target nodes confirmed suspended before each: 4/4 opened PipeWire, 0 fell back. Before: 0/4. Build clean, ctest 1019/1019, gate unchanged at 165 / 8135 with the PipeWire backend at 0, headless self-test passes.

The notice paints correctly but sits under the bank selector in a banked session until #578 lands. That is pre-existing and also hides the no-input line from #554. #578 has the fix; whichever of the two merges second needs a small rebase in `TransportBar.cpp`.

## Summary by CodeRabbit

* **New Features**
  * Added a transport-bar notice when startup cannot use the preferred audio backend and falls back to another available backend.
  * The notice identifies both the unavailable and active backends and clears after selecting a device.
* **Bug Fixes**
  * Improved PipeWire startup reliability by waiting for the audio graph to report its active quantum.
  * Updated the displayed test count from 1034 to 1039.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1034 Catch2 test cases across 193 test source files. Linux
+The C++ suite declares 1038 Catch2 test cases across 193 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
@@ -124,7 +124,7 @@ src/
   session/     # Session model + JSON serialisation
   ui/          # MainComponent, ConsoleView, channel/aux/master strips, mastering view
   util/        # Native log storage + CrashHandler signal reports
-tests/         # 1034 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
+tests/         # 1038 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
 packaging/     # .desktop, AppStream, MIME, macOS bundle — for tarball + DMG builds
 DuskStudio.md  # authoritative product spec
 MANUAL.md      # end-user manual (Pandoc-buildable to PDF via docs/build-pdf.sh)
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1038 Catch2 test cases across 193 test source files. Linux
+The C++ suite declares 1039 Catch2 test cases across 193 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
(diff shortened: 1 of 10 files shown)


marc-korte:

PR #580, opened 2026-09-10
fix(pipewire): follow PipeWire's own default sink and source

### LINKED ISSUE #577: [Audio/Linux] The PipeWire default device ignores PipeWire's own default sink and source
Found while diagnosing #576, which is a different bug in the same first-launch path. This one is #536's step 1: does first launch pick a sane audio device.

## What happens

`PipeWireAudioIODeviceType::getDefaultDeviceIndex` (`src/engine/pipewire/PipeWireAudioIODeviceType.cpp`) takes the first entry in enumeration order that is not a Monitor and not HDMI:

```cpp
const auto& names = forInput ? inputNames : outputNames;
for (int i = 0; i < (int) names.size(); ++i)
{
    const auto& n = names[(size_t) i];
    if (dusk::text::containsIgnoreCase (n, "Monitor")) continue;
    if (dusk::text::containsIgnoreCase (n, "HDMI"))    continue;
    return i;
}
```

Enumeration order is registry order, which is roughly the order the nodes appeared in the graph. PipeWire knows perfectly well which device the user considers default and publishes it in the settings metadata, and nothing reads it.

## Concrete example

This box, `pw-dump`, nodes in registry order:

| id | class | description |
|---|---|---|
| 43 | Audio/Sink | UMC1820 Multichannel |
| 44 | Audio/Source | C920 PRO HD Webcam Analog Stereo |
| 64 | Audio/Source | UMC1820 Multichannel |
| 65 | Audio/Sink | Built-in Audio Analog Stereo |
| 66 | Audio/Source | Built-in Audio Analog Stereo |
| 185 | Audio/Sink | Polaris 22 HDMI Audio Digital Stereo |

`pw-metadata`:

```
default.audio.sink   = alsa_output.pci-0000_00_1f.3.analog-stereo   (Built-in Audio Analog Stereo)
default.audio.source = alsa_input.usb-046d_HD_Pro_Webcam_C920_...   (C920 PRO HD Webcam)
```

First launch picks **UMC1820 Multichannel** for output, because it happens to be node 43. The system default sink is the built-in analog. A user whose sound comes out of their laptop speakers everywhere else opens Dusk Studio and it is routed to an interface that may not even be powered on, with nothing said.

The input pick lands on the C920 by the same accident, which agrees with `default.audio.source` here only by coincidence.

## Expected

The same device the rest of the desktop uses, unless the user has saved a choice. That is what `default.audio.sink` / `default.audio.source` mean.

## Fix shape

Read the `settings` metadata object during `scanForDevices` (the registry scan already runs a roundtrip, so it is the same trip: bind the metadata global whose `metadata.name` is `default`, take `default.audio.sink` and `default.audio.source`, both JSON objects of the form `{"name":"<node.name>"}`) and resolve those names against `outputIds` / `inputIds`. Keep the current Monitor/HDMI-skipping order as the fallback for when the metadata is absent or names a node that is gone, which is what a stripped session or a non-WirePlumber setup looks like.

A saved selection already takes precedence over the default index, so this only changes the first launch and an explicit "revert to default".

## Not the same as #576

#576 is the open failing outright on a cold device and falling back to ALSA. This is which device gets chosen when the open succeeds. Fixing #576 makes this one more visible rather than less, because PipeWire now opens on first launch and the questionable pick is what the user actually hears.

## Scope

Linux/PipeWire only. The ALSA backend has its own heuristic and is not in question here.

marc-korte:
Closes #577
Found on #536 step 1.

`PipeWireAudioIODeviceType::getDefaultDeviceIndex` took the first non-Monitor non-HDMI node in registry order and ignored PipeWire's own defaults. On the dev box that put the UMC1820 in front of the built-in analog output the desktop uses.

## Change

- During the existing scan roundtrip the registry callback binds the `default` metadata object and listens for `default.audio.sink` and `default.audio.source`. A second `pw_core_sync` gives the property events a roundtrip. The scan is marked complete before the metadata wait, so a lost second roundtrip only loses the hint, never the node list.
- `indexOfMetadataDefault` in the new `PipeWireDefaultDevice.h` is pure: parses the `{"name":...}` value, resolves against the scanned nodes, returns -1 for absent, malformed, unknown or monitor.
- The effective `default.audio.sink` is used, not `default.configured.audio.sink`. On the dev box the configured key names a Bluetooth sink that is not in the graph.
- Monitors stay excluded even when named as the default. A DAW whose record input defaults to the loopback of its own output produces takes full of whatever else was playing, with nothing on screen to say why.
- Fallback is the previous order when the metadata is absent.

## Verification

Four Catch2 cases, 12 assertions. Build clean, ctest 1018/1018, gate unchanged with all three PipeWire files at 0, headless self-test passes. Fresh-HOME launch on Xvfb now opens `Built-in Audio Analog Stereo` where it opened `UMC1820 Multichannel` before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ GPL source on this repo, so building it yourself costs you nothing but compile t
 | macOS DMG (unsigned, ad-hoc) | Working (CI publishes to private releases repo on tag) |
 | Deeper a11y (full screen-reader labels + keyboard-only mixer nav) | Floor only |
 
-The C++ suite declares 1034 Catch2 test cases across 193 test source files. Linux
+The C++ suite declares 1038 Catch2 test cases across 194 test source files. Linux
 (amd64 + arm64) and macOS builds run on every push; Windows tests run on every
 push + PR; Linux ThreadSanitizer runs on every PR + push.
 
@@ -124,7 +124,7 @@ src/
   session/     # Session model + JSON serialisation
   ui/          # MainComponent, ConsoleView, channel/aux/master strips, mastering view
   util/        # Native log storage + CrashHandler signal reports
-tests/         # 1034 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
+tests/         # 1038 Catch2 test cases declared in C++ (session, recording, MIDI, IPC, DSP)
 packaging/     # .desktop, AppStream, MIME, macOS bundle — for tarball + DMG builds
 DuskStudio.md  # authoritative product spec
 MANUAL.md      # end-user manual (Pandoc-buildable to PDF via docs/build-pdf.sh)
--- a/src/engine/pipewire/PipeWireAudioIODeviceType.cpp
+++ b/src/engine/pipewire/PipeWireAudioIODeviceType.cpp
@@ -1,9 +1,11 @@
 #include "PipeWireAudioIODeviceType.h"
 #include "PipeWireAudioIODevice.h"
+#include "PipeWireDefaultDevice.h"
 
 #include "../../foundation/Text.h"
 
 #include <pipewire/pipewire.h>
+#include <pipewire/extensions/metadata.h>
 #include <spa/utils/dict.h>
 
 #include <cassert>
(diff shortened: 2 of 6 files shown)


marc-korte:

PR #581, opened 2026-09-10
fix(ui): give the device notice a row the bank buttons do not cover

### LINKED ISSUE #578: [UI] The transport-bar notices are painted underneath the bank selector

marc-korte:
Closes #578

The transport bar centred its notices in the band between the clock and the utility cluster. `MainComponent` lays the bank buttons into the same band and they paint over it. In the default 24-strip session at any ordinary width the no-input line from #554 was never readable; only slivers of the pill showed between the buttons. The manual figure did not catch it because it was a component snapshot, which cannot contain an overlapping sibling.

## Change

The notice gets its own 22 px row at the bottom of the transport bar. The bar asks for that height only while a notice stands and `MainComponent` gives it back when it clears. The bank cluster keeps its full band. The row lives inside `TransportBar` rather than a new component because a new component would be a new coupled file and every touched file sits at its allowlist ceiling; this costs zero new tokens.

The header status label was considered and rejected: it is the session-name display and a one-shot message surface written from six places, so a persistent device notice there loses the session identity and gets overwritten by the next message.

`ScreenshotCapture` now takes the notice figure off the parent over the bar's rectangle so a sibling overlap cannot hide again. `docs/images/rec-02-no-input-notice.png` recaptured from the composed window.

## Verification

Composed-window screenshots on Xvfb with a genuinely input-less ALSA device: notice readable at 1600x900 (6 bank buttons) and 1920x1080 (3 bank buttons), layout restores with no notice, bank switching works with the notice up. Build clean, ctest 1014/1014, gate unchanged with nothing raised, headless self-test passes.


marc-korte:

Issue #575, opened 2026-09-10
[Windows] Esc does not dismiss the audio settings modal


marc-korte:
Found re-walking the demo path on a packaged build (#536), after the #558 keyboard fix.

## Setup

Windows 11 VM, no GPU. Build from `main` at 76571220, run from `%LOCALAPPDATA%\dusk-walk-2` with the Mesa `opengl32.dll` + `libgallium_wgl.dll` beside it and `GALLIUM_DRIVER=llvmpipe` in the environment, so the renderer is `llvmpipe`, not the D3D12 path that #559 covered. App window foreground, session loaded.

## Steps

1. Settings, then Settings...
2. Press Esc.

## Expected

The modal closes. MANUAL.md says so directly:

> Modal dialogs (audio settings, plugin picker, region editor, piano roll, import target picker) appear centered with a dimmed backdrop behind them. Press **Esc** or click outside the modal to dismiss.

## Actual

Nothing. The panel stays open. Pressed Esc twice on the first attempt and once on a second, fresh launch: the panel is still there in all three screenshots.

Clicking outside the panel does close it, so the modal is dismissable, just not by Esc. On the same build the panel also has no visible Close button on a 1440x900 screen: the last row is Recording offset and the panel runs off the bottom of the window, so clicking outside is the only route a user has.

## Not Linux

Walked the same step on the packaged Linux tarball from the same commit on a private Xvfb display. Esc closes the audio settings panel there first time, every time. This is Windows only.

## Also worth a look while you are in there

Transport shortcuts fire while the settings modal is open, on both platforms. Space rolls the transport and R starts a take with the panel covering the screen. That may be deliberate, so I have not filed it, but the combination of "Esc does not close it" and "R records behind it" is how I first noticed the panel was still open.

Other keyboard shortcuts are fine on this build: Ctrl+1, Ctrl+2, Ctrl+B, Ctrl+Q, Space, R and Home all work, which is what #558 was about.

PR #585, opened 2026-09-10
feat(builtin): reverb, tape echo, tape, Sunset instrument, editors, and Tape Machine 2 on the master bus

### LINKED ISSUE #75: New Feature: Built In Plugins

### LINKED ISSUE #340: Change Tape Machine to Use the new Tape Machine 2 Plugin

marc-korte:
Closes #75, closes #340.

Stage 2 of the built-in suite, on top of #583. A fresh install with nothing on the machine now has a reverb, a delay, a colour unit, a utility and an instrument, each with an editor, session state and MIDI Learn.

## The units

| id | name | core |
|---|---|---|
| `dusk.builtin.utility` | Utility | ours (#583) |
| `dusk.builtin.reverb` | Reverb | DuskVerb engine, framework-free |
| `dusk.builtin.delay` | Tape Echo | tape-echo core |
| `dusk.builtin.tape` | Tape | Tape Machine 2 core |
| `dusk.builtin.synth` | Sunset | Sunset Circuits core |

Every core is framework-free. No new `juce::` token anywhere; the gate falls from 8133 to 8106 because the editor host needed room in two strip components and that room was made by real seam migrations (value formatting off the framework string), recorded by `tools/juce-gate.sh --update`. `MainComponent.cpp` ends at its unchanged ceiling.

`BuiltinUnit::process` gained a MIDI buffer argument. Sunset decodes note on and off with velocity, aftertouch, pitch bend, mod wheel, sustain, all-sound-off and all-notes-off; the engine already sends all-notes-off to every loaded MIDI consumer on stop. Tape Echo keeps the core's fixed 4x preamp oversampling, which has no hook and would alias without it; inserts run at the strip's base rate so nothing doubles. Tape Machine 2 and Sunset fix their own internal 2x.

## Tape Machine 2 on the master bus (#340)

`MasterTape.cpp` needed no source change: every setter it calls survives with the same name and arity. Two behaviour changes ride along and are documented in MANUAL and CHANGELOG, both pinned by tests:

- The engine-wide Effect Oversampling setting no longer reaches the master tape. The core anti-aliases locally at a fixed internal rate and its oversampling parameter is inert. The setting still drives every other oversampled stage.
- Master-bus latency is a constant 56 samples where it was 0 at 1x. The tape crossfade delay tracks it. The bounce engine now trims that delay out of a render; before this it was 0 at the default factor so nothing showed.

The old JUCE-donor tape A/B test is replaced by a golden-output tone regression, because the only JUCE tape processor in the donor is v1 and nulling v2 against it would assert the opposite of what #340 is for. RMS tolerance is 0.1% (a 0.06 dB drive change trips it; peak cannot, Auto Comp hides drive in peak) and peak 0.2% as the gross-scaling guard.

## Editor

One generic ImGui `BuiltinUnitView` hosted in `DuskPanelWindow`, the same path as the audio settings panel, which renders on the software-GL Windows VM. `ParamInfo` grew section, suffix and kind with named choice tables, so machine, speed, tape type, path, EQ, reverb algorithm, echo mode, synth mode and oscillator waves show by name. A headless layout test draws every unit and asserts nothing paints outside its rectangle and every parameter is reachable; the screenshot pass on top caught a colour channel order bug and a percent sign truncating suffixes, both fixed. Opens from the slot control and the right-click menu, dim overlay, closes on Escape through the same shell branch #575 added. The aux slot opens the same modal rather than an inline embed; there is no precedent for a non-modal DGL child and inventing one here was the wrong trade. MIDI Learn: the view reports the last control moved, the slot keeps it, both strips answer with it.

Figures: `docs/images/bi-01-utility.png` to `bi-05-sunset.png`, captured through the real panel window under Xvfb.

## Donor pin

`DONOR_REV` moves in all six workflows to `a08bba29` on the donor branch `dusk-1.0-pin`, three commits on the previous pin: the Sunset Circuits core, the Tape Machine 2 core, and the two shared DSP additions Tape Machine 2 needs. Nothing else in the donor moves: no 4K-EQ-2, no Multi-Comp 2, no saturation change. The one non-additive shared change, the oversampler ring going from 64 to 128 entries, was proven byte-identical on the pinned 4K EQ and tape-echo cores. The previous pin is not an ancestor of donor main, which is why this is a branch rather than a fast-forward.

## Docs

MANUAL section per unit with controls, ranges, defaults and figure, the editor and MIDI Learn, the two master-tape changes. RELEASE-1.0.md moves #340 to 1.0 and softens the DAF bullet to consumed-and-changed-only-when-a-1.0-feature-needs-it. CHANGELOG entries. Tape formulation names are described, not named, per the manual's existing practice.

## Verification

Build clean, no new warnings. ctest 1060/1060. `tools/juce-gate.sh` 165 files / 8106 uses. Headless self-test 15/15. Verified locally against the new pin; CI reproduces it from this PR on.

## Review follow-ups

- Master tape on the Thru path: the core reports 0 latency and passes samples through bit-exact, but `MasterBus` cached the Repro figure at prepare and delayed its dry by it, so a path change after prepare put the wet 56 samples early against the reported latency and a bounce dropped its first 56 samples. Measured with an impulse before the fix. `MasterTape` now resolves the processing path's figure once at prepare and holds it, and `MasterBus` emits the aligned dry on a passthrough path, which is bit-identical to a delayed Thru. New test `master_tape_alignment.cpp` asserts arrival equals the report on the bit-exact paths.
- Unit test helpers fail on an unknown parameter id instead of returning -1 into a bounds-checked no-op.
- MANUAL's Effect oversampling paragraph no longer claims tape saturation follows the setting.
- The built-in editor window is reset explicitly with its siblings in both strip destructors, matching the compressor editor. The claimed use-after-free does not occur (`DuskImGuiHost` clears its callbacks before teardown and the hider restores in its own destructor), so this is consistency only.
- `M_PI` in the tape unit test replaced by a local constant for MSVC.


marc-korte:
@coderabbitai review

PR #599, opened 2026-09-14
Fix Retina button label scaling

### LINKED ISSUE #590: [UI/macOS] ImGui button labels are drawn at twice their size on Retina displays
Found walking the demo path on the packaged macOS build (#536), step 3: first launch.

## What happens

On a Retina Mac (backing scale 2.0), several ImGui buttons draw their labels at about twice the intended size. The button boxes are the right size, so the text crowds them or spills out.

- Startup picker: the RECENT / OPEN / NEW tabs and the Quit / Open buttons. The headings in the same view ("Recent Sessions", "No recent sessions yet.") are the right size.
- Settings: "Rescan devices", "MIDI Bindings..." and "Run Self-Test..." overflow their boxes.

Linux runs at scale 1 and looks right, which is why the reference figures never showed it.

## Cause

DAF-Widgets `textButton` treats `ButtonStyle::fontSize` as a design size and scales it itself (`const float size = ctx.s(style.fontSize);` in `opengl/DuskWidgets.cpp`). `text()` takes a size already in pixels. These Dusk call sites pass a size already multiplied by the scale into `textButton`, so it gets scaled twice:

- `src/ui/imgui/StartupView.cpp`: the tab, Open, Quit and template buttons (and the template button rounding).
- `src/ui/imgui/ChannelCompView.cpp`: the ON, Opto/FET/VCA, Soft Knee and Classic buttons.
- `src/ui/imgui/VirtualKeyboardView.cpp`: the channel and octave step buttons.
- `src/ui/imgui/PanelControls.cpp`: `formButton` passes `font->FontSize`, which is already in atlas pixels. That covers every Audio Settings form button.

`MasteringEqView.cpp` does it right (`style.fontSize = 9.0f`). A DAF re-pin would not help; `textButton` has not changed since it was added.

## Fix

Pass design sizes at those sites (`font->FontSize / ctx.scale` in `formButton`). Cover it with a headless draw-list test at `ctx.scale = 2` that measures the glyph quad height of a `formButton` and a StartupView tab. The existing ImGui view tests all run at scale 1, so running them at scale 2 as well would catch the next one.

iibaranov-IG:
Fixes #590.

`DuskWidgets::textButton` treats `ButtonStyle::fontSize` and `rounding` as design units and scales them through the widget context. Several Dusk call sites supplied values that were already scaled, so Retina displays applied the 2x scale twice.

This change passes design sizes from StartupView, ChannelCompView, and VirtualKeyboardView. `formButton` converts the selected atlas font size back to design units before passing it to the widget.

Validation:
- `tools/juce-gate.sh`
- `git diff --check`
- source-level regression assertions covering every changed call site

The full local build could not be configured in this checkout because the repository's required sibling JUCE and donor plugin sources are not present. This draft leaves the repository CI to run the complete macOS and Linux builds.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected button text sizing across channel, startup, virtual keyboard, and control panels so labels render consistently at different UI scale settings.
  - Standardized widget styling so button fonts and template-row corner rounding use consistent design dimensions without unintended scaling.
  - Improved visual consistency across interface controls when changing display scaling.


marc-korte:
@iibaranov-IG Thanks for you help!

Issue #594, opened 2026-09-11
[Takes] An overdub that partly covers an older take drops the covered part from take history


marc-korte:
Found walking the demo path on the packaged macOS build (#536), step 5: overdub.

## What happens

Three takes on one track, each recorded over the last:

| Take | Length and position | Region afterwards | Region's take history |
|---|---|---|---|
| 1 | 7.2 s at 0 | covered | n/a |
| 2 | 5.3 s at 0 | cut down to its first 0.15 s when take 3 went over it | the first 5.3 s of take 1 |
| 3 | 9.7 s from 0.15 s | 9.7 s | only the tail of take 1 (source offset 234944, 84544 samples) |

The 5.2 s of take 2 that take 3 covered is in no region's history. Delete take 3 and a 0.15 s sliver of take 2 is all that is left; nothing in the UI brings take 2 back. Cmd+Z does restore it, and the WAV is still on disk, so nothing is lost from disk, only from the session.

## Expected

When an overdub cuts into an existing region, the part it covers goes into the new region's take history, so it can be swapped back in the way MANUAL.md describes ("right-click the region and pick a previous take").

## Test

A session-level test that records over a partial overlap and checks the new region's history covers every region it trimmed.

marc-korte:
Moved to 1.1. Rather than patching the flat take history, 1.1 replaces it with Logic-style take folders: every take stays complete, and the takes that play are chosen by comping in the region editor. A partial overdub can then never drop covered audio. For 1.0, MANUAL.md documents the limit (use Undo right after recording).

PR #600, opened 2026-09-15
1.0: Stop returns to play start, zoom kept while recording, brackets arm loop and punch

### LINKED ISSUE #591: [Transport] Stopping a take leaves the playhead at the end, and the documented rewinds do not happen
Found walking the demo path on the packaged macOS build (#536), step 4: record, stop, play back.

## What happens

Record a take and press Space or Stop. The playhead stays where recording stopped. Pressing Stop again does nothing. Getting back to the start of the take takes Home or `.`.

## What a first-time user expects

Stopping after a take puts you back where the take started, ready to listen. The docs promise a rewind too, and neither holds on a default install:

- QUICKSTART.md step 4: "Press stop again to return to the start, then play to hear it back." A second Stop is a no-op, because `AudioEngine::stop()` returns early when the transport is already stopped.
- MANUAL.md, the transport bar table and the Stop bullet: "Halts playback or recording, returns the playhead to bar 1." That is only true when Settings > Playhead on Stop is "Return to start". The default is "Stay where it is (pause)" (`appconfig::getStopBehavior`).

## Proposed fix

- Add a Playhead on Stop choice "Return to where play or record started" and make it the default.
- Make Stop while already stopped rewind to zero, so the quickstart line is true.
- Correct MANUAL.md's two Stop descriptions to match.

Default to be confirmed before building.

### LINKED ISSUE #592: [Record] Punch in and out set from the ruler menu do not arm punch, so the take records the whole pass
Found walking the demo path on the packaged macOS build (#536), step 5: overdub.

## What happens

Setting the brackets with the ruler's right-click items "Set punch in here" and "Set punch out here" places the brackets but leaves punch off. Recording then captures the whole pass. In the walk, a take meant to land between 1.61 s and 2.64 s came out 9.7 s long starting at 0.15 s. The autosave written right after it has the brackets set and `punch_enabled: false`.

Drag-selecting on the ruler and choosing "Set punch in / out here" does arm punch, and punch then works exactly. A take recorded from before the brackets to well past them was written as 1.77 s placed at 5.738 to 7.509 s, the bracket range.

## Reading

In `src/ui/TapeStrip.cpp` the two separate items only call `setPunchRange`. The combined item also calls `setPunchEnabled (true)`. Nothing on screen says the brackets are set but punch is off.

## Proposed fix

The separate items arm punch once the range is valid (in before out), matching the combined item. Worth showing the brackets dimmed while punch is off, so the state can be read at a glance.

### LINKED ISSUE #593: [Timeline] The tape strip zooms out to a minute when recording starts and snaps back on stop
Found walking the demo path on the packaged macOS build (#536), steps 4 and 5.

## What happens

With a short session fitted to the window, pressing record zooms the tape strip out to at least a minute. Pressing stop snaps it back to the previous zoom. The view jumps twice on every take, and while recording the take being made is a thin sliver at the left.

## Reading

`TapeStrip::pixelsPerSecond` ignores `userZoomFactor` while `isRecording()` and returns the fit value, whose budget is `max (60 s, content * 1.2)`. The comment says this keeps the growing take on screen.

## Fix

Keep the user's zoom while recording and scroll to follow the playhead, paging when it reaches the right edge, instead of changing the zoom.

marc-korte:
Rebuilt from the unpushed `fix/take-transport-walk` worktree branch onto current `main`, without its take-history commits.

## What's in it

- **Stop returns to where play or record started** (closes #591). New Playhead on Stop choice, now the default; saved choices keep their meaning. Stop while already stopped rewinds to the start, so QUICKSTART's "press stop again" holds. MANUAL's Stop descriptions match.
- **Placing loop and punch brackets arms them** (closes #592). Ruler menu items, `Shift+[` / `Shift+]` and `[` / `]` arm punch or loop once in < out; brackets draw hollow while their mode is off. This replaces the inline punch arming `main` got in `1026f93` with the shared `armPunchForRange()`.
- **Zoom stays put while recording, and the view pages to follow the playhead** (closes #593). Paging decision is a pure helper with tests (`TimelineFollow.h`, `tests/timeline_follow.cpp`).
- **MANUAL: 1.0 take history limits.** A partial overdub does not keep the covered part in history, and Delete removes the whole stack; Undo recovers both.
- README test counts updated.

## Left out on purpose

`a420d56`, `f9bc72c`, `ba181c2` (#594/#595 take history patches) stay on `fix/take-transport-walk` only. Take history is being replaced by Logic-style take folders in 1.1, so #594/#595 moved to the 1.1 milestone.

## Verification

Local test build, full Catch2 suite, JUCE gate and app build are running; results to follow in a comment.

## Summary by CodeRabbit

* **New Features**
  * Added a “Return to where play or record started” option for Playhead on Stop, now the default.
  * Loop and punch brackets now arm automatically when the in point precedes the out point.
  * Added clear visual states for active and inactive loop/punch brackets.

* **Bug Fixes**
  * Recording now preserves zoom and follows the playhead without unexpectedly zooming out.
  * Updated stopping and seeking behavior for more consistent playback navigation.

* **Documentation**
  * Updated the manual, quickstart guide, changelog, and keyboard reference to reflect these workflow changes.


marc-korte:
Local verification on macOS arm64 (head 7a9d4ad; donor checkout at 92977f85, one pin newer than this branch's DONOR_REV, which CI uses):

- `dusk-studio-tests` build: no errors
- Full Catch2 suite: 1067 / 1069 passed. The 2 failures, `delay unit turns silence into silence` and `delay unit at its defaults is unity gain`, also fail on main locally and are not touched here.
- `tools/juce-gate.sh`: pass (165 coupled files, 8100 uses)
- App build: no errors

